PR #43, opened 2018-04-04
Added messages to the single resource bundle file and a writeInputStr…

mwooten:
Hi Carolyn,

Can you review/approve the changes to two files I had to change in weblogic-deploy-tooling, because I need to use code from weblogic-deploy-tooling in the new weblogic-deploy-testing project I'm working on?

The details are in the comments of Issue# 41 and Issue # 42


core/src/main/resources/oracle/weblogic/deploy/messages/wlsdeploy_rb.properties
```diff
@@ -1115,3 +1115,42 @@ WLSDPLY-20019=Filter entry in {0} has neither ID or path
 WLSDPLY-20020=Filter ID {0} is invalid
 WLSDPLY-20021=Filter path {0} does not exist
 WLSDPLY-20022=Error loading filter path {0}
+
```
ddsharpe: Why are integration test messages in WDT?  Shouldn't these be in the integration test project?

mwooten: It looks like the stuff we found with ExceptionHelper.java (and exception_helper.py) using a static resource bundle name, is the reason I was getting the MissingResourceException. I will find a way to not use the get_message() function in either of those, in the weblogic-deploy-testing project. 

The changes made to wlsdeploy_rb.properties can be reverted, but I still need the change to FileUtils.java to be reviewed and merged.

PR #59, opened 2018-04-18
Added a alias JSON file for Log as a root folder

### LINKED ISSUE #57: Need to add an alias JSON file for Log
There needs to be a **Log.json** file, because it is possible that someone would want to control the name and attributes of the domain log file. It would be similar to how we have a JTA.json and JMX.json files, for managing the configuration of those, when they are root folders in the topology section of a model.

mwooten:
Fixes #57 

But it did surface a different issue. What if a root folder is only available in ONLINE mode? That's the case with this /Log folder...but we don't currently have anything at the folder level in the alias JSON files, for specifying that. We have "wlst_mode": ["online", "offline" or "both"] at the attribute level, but not at the folder level.


core/src/main/resources/oracle/weblogic/deploy/aliases/category_modules/Log.json
```diff
@@ -0,0 +1,51 @@
...
+        "StdoutFilter":                                  [ {"version": "[10,)",     "wlst_mode": "both", "wlst_name": "StdoutFilter",                                  "wlst_path": "WP001", "value": {"default": "None"  }, "wlst_type": "string" } ],
+        "StdoutFormat":                                  [ {"version": "[10,)",     "wlst_mode": "both", "wlst_name": "StdoutFormat",                                  "wlst_path": "WP001", "value": {"default": "standard"}, "wlst_type": "string" } ],
+        "StdoutLogStack":                                [ {"version": "[10,)",     "wlst_mode": "both", "wlst_name": "StdoutLogStack",                                "wlst_path": "WP001", "value": {"default": "true"  }, "wlst_type": "boolean"} ],
+        "StdoutSeverity":                                [ {"version": "[10,)",     "wlst_mode": "both", "wlst_name": "StdoutSeverity",                                "wlst_path": "WP001", "value": {"default": "Notice"}, "wlst_type": "string" } ]
+    },
+    "wlst_attributes_path": "WP001",
+    "wlst_paths": {
+        "WP001": "/Log/%DOMAIN%"
```
ddsharpe: Shouldn't this be /Log${:s}/%DOMAIN%

core/src/main/resources/oracle/weblogic/deploy/aliases/category_modules/Log.json
```diff
@@ -0,0 +1,51 @@
+{
+    "copyright": "Copyright (c) 2017, 2018, Oracle and/or its affiliates. All rights reserved.",
+    "license": "The Universal Permissive License (UPL), Version 1.0",
+    "wlst_type": "Log",
+    "folders": {},
+    "attributes": {
+        "BufferSizeKb":                                  [ {"version": "[10,)",     "wlst_mode": "both", "wlst_name": "BufferSizeKb",                                  "wlst_path": "WP001", "value": {"default": 8       }, "wlst_type": "integer"} ],
```
ddsharpe: Are these available offline?  If not, shouldn't the wlst_mode be offline only?

mwooten:
Set wlst_mode to "online" of each attribute. This makes the /Log root folder only visible in WLST online mode, which is the case in WLST proper.

core/src/main/resources/oracle/weblogic/deploy/aliases/category_modules/Log.json
mwooten: Nope. It (the /Log folder) is only visible in WLST in ONLINE mode, and then the folder name is Log...not Logs

CarolynRountree:
Log is available in offline.
Ran this through my tester, and found case is wrong for online attribute: It should be BufferSizeKB
And this offline online attribute, ServerLoggingBridgeAtRootLoggerEnabled, is missing.

mwooten:
Changed wlst_name for BufferSizeKb to BufferSizeK${b:B}. Both FileTimeSpanFactor and ServerLoggingBridgeAtRootLoggerEnabled are only available in OFFLINE, so I made that change, as well.

ddsharpe:
Its VERY rare for something to be in Offline and not in Online.  Are you positive?

PR #103, opened 2018-05-19
Added missing alias_helper methods and some unit tests

### LINKED ISSUE #102: Aliases method missing from alias_helper.py
It looks like the get_model_resources_top_level_folder_names() method is missing from the alias_helper.py file. This is causing a Python stack trace when the -print_usage option on validate is invoked.

mwooten:
Fixes #102 

Actually, a couple methods were missing and probably went unnoticed because Python isn't a compiled programming language. I added 4 unit tests to validation_test.py in order to combat this type of thing with print usage, in the future

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/util/alias_helper.py
+++ b/core/src/main/python/wlsdeploy/tool/util/alias_helper.py
@@ -522,6 +522,20 @@ class AliasHelper(object):
         """
         return self.__aliases.get_model_topology_top_level_folder_names()
 
+    def get_model_resources_top_level_folder_names(self):
+        """
+        Get the model resources top-level folder names.
+        :return: the list of top-level resources folders
+        """
+        return self.__aliases.get_model_resources_top_level_folder_names()
+
+    def get_model_app_deployments_top_level_folder_names(self):
+        """
+        Get the model app_deployments top-level folder names.
+        :return: the list of top-level app_deployments folders
+        """
+        return self.__aliases.get_model_app_deployments_top_level_folder_names()
+
     def get_model_attribute_names(self, location):
         """
         Get the model attribute names.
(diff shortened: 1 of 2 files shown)


core/src/main/python/wlsdeploy/tool/util/alias_helper.py
```diff
@@ -522,6 +522,20 @@ def get_model_topology_top_level_folder_names(self):
         """
         return self.__aliases.get_model_topology_top_level_folder_names()
 
+    def get_model_resources_top_level_folder_names(self):
```
ddsharpe: Why is it necessary to add these methods to the helper?  Isn't it sufficient to access the underlying alias method directly from the unit test?  I don't see the reason to expose this as a "public" API.

mwooten: I'm confused. 

First, all of the other public methods on the aIiases.py are exposed on the alias_helper.py, so I assumed that that was the design intention of the helper. The way RP explained it, the xxx_helper was there so the caller wouldn't have to have a try/except around every call they make. 

Second, I didn't do this code change because a unit test was bombing out. As I stated in the comment for the issue, I did the code change because the usage_printer.py is bombing out due to the missing helper methods. I think you might be surmising that since usage_printer.py isn't one of the files I needed to change but a unit test is, that this change isn't really needed. The unit test file was changed because the existing unit tests weren't written in a way to catch the error, in the first place.

CarolynRountree:
alias_helper actually still throws an exception. The helper was created so that all of the different tools would have the same formatted exception message instead of each tool formatting its own different type of message. And this would allow the tools to allow the exceptions to percolate all the way up to the top without doing a catch on the exception on each call, since the error message was already written to the log. Limited value though.

Issue #108, opened 2018-05-23
validator no longer using variable_file argument on call to validate


CarolynRountree:
I just merged master into my WIP branch for injecting variables into the model. Code has changed in the validator that ignores the variable file name passed on the function call to validate. My work is blocked by this problem.

Here is the new code that ignores the variable file name argument:
        if variables_file_name is not None:
            self._logger.info('WLSDPLY-05004', variables_file_name, class_name=_class_name, method_name=_method_name)
            try:
                if self._model_context.get_variable_file():
                    self._variable_properties = variables.load_variables(self._model_context.get_variable_file())
                variables.substitute(model_dict, self._variable_properties)
            except VariableException, ve:
                ex = exception_helper.create_validate_exception('WLSDPLY-20004', 'validateModel',
                                                                ve.getLocalizedMessage(), error=ve)
                self._logger.throwing(ex, class_name=_class_name, method_name=_method_name)
                raise ex

mwooten:
The fix for this is in the second commit for Pull Request#109

PR #109, opened 2018-05-23
Updated WLST data type compatibility check method to see "class java.…

### LINKED ISSUE #107: Fix compatibility data typing for Python long
validation_utils.is_compatible_data_type() method needs to treat WLST type "long", as being equivalent to java.lang.Long Java type and <type 'javainstance'> Python type.

mwooten:
Fixes issues #107, #108 and #112


core/src/main/python/wlsdeploy/tool/validate/validator.py
```diff
@@ -108,9 +109,11 @@ def validate_in_standalone_mode(self, model_dict, variables_file_name=None, arch
         """
         _method_name = 'validate_in_standalone_mode'
 
+        cloned_model_dict = copy.deepcopy(model_dict)
```
ddsharpe: Please add a comment here to describe why its necessary to do a deep copy so someone doesn't accidentally change it back.

PR #159, opened 2018-09-06
fix problem with boolean types in securityconfiguration

CarolynRountree:
The boolean type changed in aliases


CarolynRountree:
I spot checked this attribute in other authentication providers. However, it is possible several of the boolean values are incorrect in the alias definitions. This will cause create and deploy to fail.

Issue #122, opened 2018-05-30
Reduce chatter of alias and version related logstreams


CarolynRountree:
There are a few log messages, mostly entering() and exiting() that fills any tool log file with messages, making it difficult to use the logs to troubleshoot. I would like to suggest that we either remove these entering and exiting logs completely, or replace with a single log at finest level. Also, there is one log that could be changed from finer to finest.

####<May 30, 2018 11:26:29 AM> \<FINER\> \<VersionUtils\> \<compareVersions\> <> <ENTRY 12.2.1.3.0 10>
####<May 30, 2018 11:26:29 AM> \<FINER\> \<VersionUtils\> \<compareVersions\> <> <RETURN 1>

####<May 30, 2018 11:26:29 AM> \<FINER\> \<VersionUtils\> \<isVersionInRange\> <> <ENTRY 12.2.1.3.0 [10,)>
####<May 30, 2018 11:26:29 AM> \<FINER\> \<VersionUtils\> \<isVersionInRange\> <> \<RETURN true\>

####<May 30, 2018 11:26:29 AM> \<FINER\> \<VersionUtils\> \<getLowerAndUpperVersionStrings\> <> \<ENTRY [10,)\>
####<May 30, 2018 11:26:29 AM> \<FINER\> \<VersionUtils\> \<getLowerAndUpperVersionStrings\> <> \<RETURN [Ljava.lang.String;@24203248\>

####<May 30, 2018 11:26:29 AM> \<FINER\> \<AliasEntries\> \<__resolve_attribute_by_wlst_context\> <> \<ENTRY Server/JTAMigratableTarget NumberOfRestartAttempts\>
####<May 30, 2018 11:26:29 AM> \<FINER\> \<AliasEntries\> \<__resolve_attribute_by_wlst_context\> <> \<RETURN\>

####<May 30, 2018 11:26:29 AM> \<FINER\> \<AliasEntries\> \<__resolve_attribute_by_wlst_context\><> \<Testing Server/JTAMigratableTarget/NonLocalPostAllowed with [10,)\>
####<Jun 1, 2018 8:59:23 AM> \<FINER\> \<AliasEntries\> \<get_dictionary_for_location\> <> \<ENTRY model_folders = ['SelfTuning', 'WorkManager'],  'name_tokens' = {'SELFTUNING': 'NO_NAME_0','DOMAIN': 'SingleAppDomain','WORKMANAGER': 'CtxReqWM'}\>
[...]

PR #160, opened 2018-09-07
modifications for log chatter

CarolynRountree:
Please do not delete the repository or close the corresponding issue for this pull request until you have read the below:

I changed the logging properties to set the aliases and versions loggers to level FINE. This will reduce the ENTER, EXIT chatter. We can request the end user to change those logger levels if needed. However, the quality of logs in the aliases is not ideal. Informative loggers at the FINE level should be added to the aliases code. I have yet to find the information I need for an alias issue from the entering,exiting code. Entering,Exiting loggers is about all that aliases logs except for those at INFO and above.

If setting the version and alias loggers to level FINE is acceptable, we can close the issue and delete the repository. I suggest we open another issue for quality of logs in aliases if we need to spend our time on this. Else, we should add loggers as we find the need when troubleshooting issues.

I put loggers into the logging.properties for the more import logger names. This will be helpful to the user since the logger name is not included in the log stream. The user does not know the names or where to in the source to find the names.

The WDT logger names I did not include in the logger properties are as follows:
wlsdeploy.encrypt
wlsdeploy.validate
wlsdeploy.tool.util
wlsdeploy.json
wlsdeploy.archive
wlsdeploy.yaml
wlsdeploy.model
wlsdeploy.translator
wlsdeploy.variables

### DIFF
--- a/core/src/main/java/oracle/weblogic/deploy/aliases/VersionUtils.java
+++ b/core/src/main/java/oracle/weblogic/deploy/aliases/VersionUtils.java
@@ -341,7 +341,7 @@ public final class VersionUtils {
             LOGGER.throwing(CLASS, METHOD, ve);
             throw ve;
         }
-        LOGGER.exiting(CLASS, METHOD, result);
+        LOGGER.exiting(CLASS, METHOD, Arrays.toString(result));
         return result;
     }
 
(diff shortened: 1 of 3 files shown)


CarolynRountree:
Issue #122

PR #227, opened 2019-01-25
Fix Restricted JRF template creation and blacklist issue, also use as…

jshum2479:
…sign cluster to server instead of set in wlst_helper since using set will confuse targetting for templates apps/libs.


jshum2479:
remove after merge

PR #245, opened 2019-02-14
Updated README with TOC

ddsharpe:


ddsharpe:
Agreed, we need a section for how to create a model from scratch.  The current doc is lacking in this topic.  The Variable Injection section matches the documentation that existed before this change, as well as the name of the tool.  I am not sure why the TOC title should not match.

As for the encrypt tool, validate tool, and injection tool, they kind of fit where they are but not really.  But, they aren't a perfect fit for the Model section either.  I am going to push the changes as is, but I think there is room for improvement with the organization of these model-utility-like tools.

PR #248, opened 2019-02-19
review and edits

rosemarymarano:
Mostly minor edits. 

In README file, aligned the headings to correspond with the TOC: three headings at the primary level (head 2) and the rest at the secondary level (head 3). If that's not what you want, we can change the headings in the doc and TOC to reflect your desired structure.


README.md
```diff
@@ -71,15 +71,15 @@ The model is written in YAML (or optionally, JSON).  The YAML parser, built into
...
+### Top-Level Model Sections
+The tooling has four top-level model sections:
 
 - `domainInfo`     - The location where special information not represented in WLST is specified (for example, the libraries that go in `$DOMAIN_HOME/lib`).
 - `topology`       - The location where servers, clusters, machines, server templates, and other domain-level configuration is specified.
-- `resources`      - The location where resources and services are specified (for example, data sources, JMS, WLDF)
+- `resources`      - The location where resources and services are specified (for example, data sources, JMS, WLDF).
 - `appDeployments` - The location where shared libraries and applications are specified.
 
-##### Simple Example
```
ddsharpe: Why does it matter if these are H3 or H5 markers?  I injected this header so that I could create a TOC link to it, since there were no "Figure" or "Code" markers that I could find.  This header and the Top-Level Model Sections were both added for the same reason and weren't really intended to be anything other than page references.

rosemarymarano: Hmmm, I understand. I can revert them to H5s. Would you consider having their links indented in the TOC? I suggest this because the TOC structure typically mimics the doc structure. Considering that some links in this TOC go to other docs (in the site dir) we don't have to follow this style guideline exactly but it makes sense to me to have subordinate paras at subordinate levels in the TOC. If you agree, I'll revert my heading changes to the doc and show you what the indented TOC would look like.

ddsharpe: Yes, I am open to the indented links.  I agree, thanks.

rosemarymarano: The TOC stylesheet wouldn't allow me to indent more than one step when going from the head 2 "The Model" to a head 5 "Top-Level Model Sections," so I put the headings and TOC back to the way you had it.

BTW, the code you have under Simple Example and Model Names is marked as ```yaml and thus appears with a particular coloration and such. However, the code under Model Semantics and Modeling Security Providers is marked as ```$yaml and has no special effects. Is this as you wanted it?

ddsharpe: good catch, thank you.  No, that is not the intent.  Strangely enough, the same user put both of those in about a week apart.  None of them should be $yaml

PR #251, opened 2019-02-22
Issue#168 arrays needed to version folders

CarolynRountree:
If a 'Provider' is found in name for providers in 11g during offline discover, WARNING and the entire SecurityConfiguration is removed from the model.

This final a redesign at request of reviewer. The original used a get_method in the new folder level array. I left in fixes that I found for the folder level array.


CarolynRountree:
I do not know why the wercker pipeline failed - all unit tests fun

core/src/main/python/wlsdeploy/tool/discover/discoverer.py
```diff
@@ -631,6 +636,24 @@ def wlst_cd(self, path, location):
...
+        Perform any special handling for the folder or folder names.
+        :param location: current context of location
+        :return: Original name or processed name value
+        """
+        return self._inspect_security_folder_name(folder_name, location)
+
+    def _inspect_security_folder_name(self, folder_name, location):
+        # This is so clunky. My alias definition change was rejected.
+        if (not self._weblogic_helper.is_version_in_12c()) and self._wlst_mode == WlstModes.OFFLINE and \
+                'SecurityConfiguration' in location.get_folder_path() and 'Provider' == folder_name:
```
rakillen: The location check should be in aliases, maybe is_security_provider(location)?  Rather than checking folder path, the implementation could be is_artificial_type_folder() for now, and expand if other artificial types turn up.

core/src/main/python/wlsdeploy/aliases/aliases.py
```diff
@@ -94,6 +94,13 @@ def get_mode_enum(self):
         """
         return self._wlst_mode
 
+    def get_version(self):
```
rakillen: I don't think this is called anywhere?

PR #270, opened 2019-03-01
Jira WDT-3 - Improve error messaging when domain read fails due to…

rakillen:
Improve error messaging when domain read fails due to WLS / domain version mismatch

### DIFF
--- a/core/src/main/python/discover.py
+++ b/core/src/main/python/discover.py
@@ -289,8 +289,9 @@ def __connect_to_domain(model_context):
         try:
             wlst_helper.read_domain(model_context.get_domain_home())
         except PyWLSTException, wlst_ex:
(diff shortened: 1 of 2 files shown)


CarolynRountree:
I imagine I would be scratching my head saying why does it think my domain is <blah blah> version?

Also, online connect could have the same problem - running WLST from a different version than where the admin server is started. And I haven't even gotten to the check that the domain home is the same version/location as the connect -- which is a different JIRA altogether

rakillen:
I think the message could be clearer, maybe "Reading the domain at {0} with Oracle home version {1} failed: {2}". Ideally we would point out the version discrepancy, but we can't determine the domain home version before it is read, and the exception thrown by WLST is not specific enough to detect this case.

PR #283, opened 2019-03-15
fix the npe caused by another option - domain_home

jshum2479:
fixed atp extract problem when domain_home is passed and use system property for oracle.jdbc.fanEnabled=false instead of OS env

### DIFF
--- a/core/src/main/python/create.py
+++ b/core/src/main/python/create.py
@@ -10,6 +10,7 @@ from java.io import IOException
 from java.lang import IllegalArgumentException
 from java.lang import IllegalStateException
 from java.lang import String
+from java.lang.System import System
 from oracle.weblogic.deploy.create import CreateException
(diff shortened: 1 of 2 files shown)


core/src/main/python/create.py
```diff
@@ -341,7 +341,9 @@ def validateRCUArgsAndModel(model_context, model):
         has_atpdbinfo = atp_helper.has_atpdbinfo(rcu_db_info)
 
         if model_context.get_archive_file_name() and not has_regular_db:
-            os.environ['oracle.jdbc.fanEnabled'] = 'false'
+            import java.lang.System as javasys
```
rakillen: PEP-8 has warning about this - can import be at the top?

jshum2479: Yes, already made the change

Issue #276, opened 2019-03-06
Domain discovery fails due to incorrectly generated output archive filename


shanedonohoe:
When running the discover domain util, I have noticed the output archive filename has been generated incorrectly. When the tool creates a temporary zip filename, the generated filename incorrectly includes the parent directories twice.

```
23790/2:        open("/home/shane/mydomain_archive.zip", O_RDONLY) = 460
23790/2:        open("/home/shane/home/shane/mydomain_archive3568716833221357580.zip", O_RDWR|O_CREAT|O_EXCL, 0666) Err#2 ENOENT
```

The above truss output was produced with the archive file argument as `-archive_file /home/shane/mydomain_archive.zip`.

It looks like `nameComponents[0]` contains the full path (minus file extension), but should only contain the filename. `createTempFile` takes the directory as a separate argument, so it is being prepended to the already full path.

https://github.com/oracle/weblogic-deploy-tooling/blob/1ff34b24b6dbb777de1cd476bba3d976e6a23fb5/core/src/main/java/oracle/weblogic/deploy/util/WLSDeployZipFile.java#L787

The problem may be below with `getFileName` returning the full path.

https://github.com/oracle/weblogic-deploy-tooling/blob/1ff34b24b6dbb777de1cd476bba3d976e6a23fb5/core/src/main/java/oracle/weblogic/deploy/util/WLSDeployZipFile.java#L778

https://github.com/oracle/weblogic-deploy-tooling/blob/1ff34b24b6dbb777de1cd476bba3d976e6a23fb5/core/src/main/java/oracle/weblogic/deploy/util/WLSDeployZipFile.java#L86-L88

jshum2479:
Can you describe the problem you are facing first ?  The lines you quoted are for temporary files only.  Are you not seeing the the archive file in your home directory after completion ?

shanedonohoe:
The problem we're facing is that domain discoveries fail when multiple artefacts need to be added to the output archive. The first artefact is added to the archive, but for subsequent artefacts the tool tries to create a temporary archive. The path of this temporary archive is being generated with the parent directories repeated (e.g. `/home/shane/home/shane/mydomain_archive3568716833221357580.zip` rather than `/home/shane/mydomain_archive3568716833221357580.zip`). We can see this from the truss output above. If we create the parent directory `/home/shane/home/shane`, the output artefacts are generated fine, but this is obviously undesirable.

The actual message is listed below, but doesn't provide as much context.

```
[...]

PR #290, opened 2019-03-18
Fix how we calculate temporary file for archive

jshum2479:
Prior than jdk 8,  java.io.File generateFile does not strip down the prefix to base name in createTempFile, we are changing how we pass in the prefix now.


rakillen:
Fixes #276

PR #325, opened 2019-03-27
Issue #298 - Add admin console alias for discover, create, update

### LINKED ISSUE #298: Unable to set attributes on AdminConsoleMBean
Doesn't look like we can set attributes on the AdminConsoleMBean?:
```yaml
topology:
    Name: mydomain
    AdminConsole:
        CookieName: mydomaincookie
```
The model fails validation:
```
<WLSDPLY-05026> <AdminConsole is not one of the folder names allowed in model location topology:/>'
```
Also, these values aren't added to the model during domain discovery.

rakillen:
Fixes #298


core/src/main/resources/oracle/weblogic/deploy/messages/wlsdeploy_rb.properties
```diff
@@ -607,7 +607,7 @@ WLSDPLY-06609=Discovering {0} Unix machines
 WLSDPLY-06610=Adding Unix Machine {0}
 WLSDPLY-06611=Discovering {0} machines
 WLSDPLY-06612=Adding Machine {0}
-
+WLSDPLY-06613=Discovering Domain Admin Console
```
ddsharpe: Shouldn't Admin Console be one word?

rakillen: Revised this somewhat, but one word now

Issue #345, opened 2019-04-04
SAF Agent creation / No target found with name of created Agent


wszczurek:
WDT: 0.22
WDT Mode: OnLine
WebLogic: 10.3.6

Usecase:
- Create SAF Agent
- Create JMS Resource and target SAF Agent

The SAF Agent is created successfully.
See file model 0-WDTSAF-agent.yaml in attachment.

Once the Agent is created the JMS Resource creation fails.
See file model 1a-WDTSAF-jmsresource-sub-error-remote.yaml in attachement.

The Error received:
```
####<Apr 4, 2019 1:19:01 PM> <SEVERE> <update> <main> <WLSDPLY-09015> <updateDomain deployment failed: No target found with name WDTSAFAgent-1>

Issue Log for updateDomain version 0.22 running weblogic version 10.3.6.0 online mode:

SEVERE Messages:

        1. WLSDPLY-09015: updateDomain deployment failed: No target found with name WDTSAFAgent-1

Total:       WARNING :     0    SEVERE :     1

updateDomain.sh failed (exit code = 2)
```

Looks like the SAFAgent is not taken into account in __find_target method.
I have work-around this by changing the attribute_setter.py.
Will provide input via merge request what helped.
After this change i was able to execute models from both files.

wszczurek:
The attachment for models.
[saf-agent-jmsres-model.zip](https://github.com/oracle/weblogic-deploy-tooling/files/3043675/saf-agent-jmsres-model.zip)

rakillen:
Fixed with PR #346

PR #346, opened 2019-04-04
Fix for issue #345 - Imported SAF_AGENT from model_constants and exte…

wszczurek:
This is my fix for issue #345.
Added importing of SAF_AGENT and use it later.
Extended __find_target method.

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/util/attribute_setter.py
+++ b/core/src/main/python/wlsdeploy/tool/util/attribute_setter.py
@@ -55,6 +55,7 @@ from wlsdeploy.aliases.model_constants import RESOURCE_MANAGEMENT
 from wlsdeploy.aliases.model_constants import RESOURCE_MANAGER
 from wlsdeploy.aliases.model_constants import RESPONSE_TIME_REQUEST_CLASS
 from wlsdeploy.aliases.model_constants import REST_NOTIFICATION
+from wlsdeploy.aliases.model_constants import SAF_AGENT
 from wlsdeploy.aliases.model_constants import SAF_IMPORTED_DESTINATION
 from wlsdeploy.aliases.model_constants import SAF_QUEUE
 from wlsdeploy.aliases.model_constants import SAF_TOPIC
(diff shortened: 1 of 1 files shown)


rakillen:
With the default value for ``include_saf`` set to True, SAF agents will be allowed as targets for other elements.
I suggest you just use the existing ``include_jms`` to determine whether to invoke the SAF agent check.
That flag is only set for the SubDeployment / Target attribute through attribute_setter.set_target_jms_mbeans().
This would expand the meaning of that flag to "include targets for JMS subdeployments", which seems OK.

wszczurek:
@ddsharpe 
I have not done it before. I have sent the signed OCA in pdf format couple of hours ago to "oracle-ca_us [at] oracle [dot] com". Hope it will be processed quickly on your side.

@rakillen 
Ok, was not sure before if I can reuse the same condition as in the chain could not find when `set_target_jms_mbeans()` is executed, as it is the only place which i could see the the `include_jms` is set to `True`.
```
[python]$ grep -rli include_jms .
./wlsdeploy/tool/util/attribute_setter.py
[python]$
```
```
[weblogic-deploy]$ grep -rli set_target_jms_mbeans .
./lib/python/wlsdeploy/tool/util/attribute_setter.py
[weblogic-deploy]$
```

I will change and test it. 
Once I will receive info from OCA - do you want me to proceed via existing Pull Request or close this one and create new one ?

ddsharpe:
Proceed with existing.  But, you will need to tag a commit as described above.

ddsharpe:
@wszczurek your OCA has been approved.  Please add the --signoff commit and finish your changes and testing.

rakillen:
Regarding ```set_target_jms_mbeans()```, the public methods in ```attribute_setter.py``` are used to set attributes that can't be handled by simple data manipulation. Each is indirectly associated one or more attributes in the alias files (such as JMSSystemResource.json). That's how I determined that ```set_target_jms_mbeans``` was used in this one place, but ```set_target_mbeans``` was used a lot. Hope that helps. We are planning to improve the documentation for this.

wszczurek:
Thanks for more information related to internals.
Definitely that would be good to have high level design and internal concept described how the tool works.

I have changed the implementation and reused the `include_jms` condition.
Signed off, squashed and pushed forcefully to my fix branch.

rakillen:
Fixes #345

PR #358, opened 2019-04-16
Wdt 228 target dynamic servers to cluster

CarolynRountree:
This pull request is for JIRA#WDT-254

Revise based on zoom discussion:

The problem: 12C only problem. How do we assign template resources to a dynamic cluster?  For configured clusters and stand-alone managed servers, WDT uses the WLST  function setServerGroups(server_name, [servergroups]). The server group list contains all "user" server groups for the set of templates in the WDT typedef.  

Current solution for createDomain and updateDomain.
1. apply the base and extension templates to the domain (only allowed in createDomain)
2. Create/update all the topology components.
3. setServerGroups to every server in the domain for the server group list in the typedef
4. Create/update all the resource and app deployments.

updateDomains does step 2 and 3 if the topology section is present in the model, and regardless of what is in the topology,  Why?  Should setServerGroups only be done  to a new stand-alone ms server or configured cluster? What other topology MBean updates/adds would require setServerGroups to run against all servers in the domain.

FYI: we have an user issue requesting WDT to allow adding template(s) in updateDomain. Sounds reasonable. What if you are installing a product to an existing domain instead of creating a new domain. In fact, the WLS wizard allows a user to do this)

WLST setServerGroups issue: For updateDomain using online WLST. when a new managed server or configured cluster is added to the the domain, the setServerGroups is run on the new entities, but the updated config.xml does not show this update.

Back to dynamic clusters. Monica noted that the upper stack products are not certified with dyanmic clusters. In fact the operator does not currently support dynamic clusters in domains with these upper stack products.

BUT - JRF resources can be targeted. This works well when creating a JRF or RestrictedJRF domain (i.e. -domain_type JRF). The JRF resources are now targeted to dynamic clusters using applyJRF. With upper stack products, such as SOA, this is an incomplete solution.

WDT will ONLY target JRF resources using applyJRF to dymamic clusters. The managed servers and configured clusters are still targeted using the setServerGroups.

Issue with applyJRF: Cannot target resources to a dynamic cluster in online WLST using applyJRF. ApplyJRF looks for the OPSS data source in online (not in offline) which, of course, does not exist in a RestrictedJRF domain. This really is only a problem when a new dynamic cluster  is added by the updateDomain and this same scenario works using offline WLST.

Another Issue with setServerGroups. The problem listed in number 4 below. Derek identified that the apps and startup class should NOT be targeted to managed servers, only to the admin server. the applyJRF does not target these resources to the dynamic cluster. However, the setServerGroups DOES target the resources to ms/clusters.

Future solution for assigning all template resources to dynamic clusters: Use the CIE template reporter API to find the resources for a server group in a template. Manually assign/target the resources to the dynamic cluster. The template reporter is  available in oracle home versions which support the select/load templates.

(Continue with the original comments)

The following situations are caveats to this scenario.f

1. WDT targets JRF resources only. At the time that upper stack products are certified with dynamic clusters and WLST has supplied a way to target extension template resources to dynamic clusters, support will be added to WDT. For now, a warning message is logged to STDOUT when encountering dynamic clusters that should be targeted upper stack products.
2. Cannot target JRF resources when running updateDomain in online WLST mode and domain_type of RestrictedJRF. The FMW applyJRF is really for targeting JRF resources only, but is used by WDT to target resources for the RestrictedJRF domain type. The applyJRF in online checks for the OPSS data source (this check is NOT made in offline mode). When this scenario is encountered, the updateDomain will stop with an exception.
3. Cannot target any resources to new configured clusters or stand alone managed server when running updateDomain in online WLST  This is because of an issue in online WLST setServerGroups. When new stand-alone managed servers or configured clusters are encountered in the model when running online, the updateDomain will stop with an exception.
4. There is a slight difference in the JRF resources that are targeted to dynamic clusters using applyJRF versus to clusters using setServerGroups. The following JRF / WSM resources are not targeted to the dynamic cluster.

```
  <app-deployment>
    <name>state-management-provider-memory-rar</name>
    <target>configured-cluster,mixed-cluster,managed-server-3</target>
    <module-type>rar</module-type>
    <source-path>C:/oracle/fmw12213/oracle_common/modules/com.oracle.state-management.state-management-provider-memory-rar-impl.rar</source-path>
    <staging-mode>nostage</staging-mode>
  </app-deployment>
  <library>
    <name>oracle.wsm.idmrest.sharedlib#1.0@12.2.1.3</name>
    <target>configured-cluster,mixed-cluster,admin-server,managed-server-3</target>
    <source-path>C:/oracle/fmw12213/oracle_common/modules/oracle.wsm.common/wsm-idmrest-sharedlib.war</source-path>
    <staging-mode>nostage</staging-mode>
  </library>
  <startup-class>
    <name>Web Services Startup Class</name>
    <target>configured-cluster,mixed-cluster,admin-server,managed-server-3</target>
    <deployment-order>150</deployment-order>
    <class-name>oracle.j2ee.ws.server.WebServiceServerStartup</class-name>
    <failure-is-fatal>false</failure-is-fatal>
    <load-before-app-deployments>true</load-before-app-deployments>
    <load-before-app-activation>true</load-before-app-activation>
  </startup-class>
```

The following scenarios have been tested:
createDomain with domain_type WLS
10.3.6 (regression only)
12.2.1.3
createDomain with domain_type RestrictedJRF
12.2.1.3
12.2.1.4
createDomain with domain_type JRF
10.3.6 (regression only)
12.2.1.3
12.2.1.4

updateDomain in offline mode with domain_type WLS
10.3.6
12.2.1.3
updateDomain in offline mode with domain_type RestrictedJRF
12.2.1.3
12.2.1.4
updateDomain in offline mode with domain_type JRF
10.3.6 (regression only)
12.2.1.3
12.2.1.4

updateDomain in online mode with domain_type WLS
(open issue in 10.3.6 in online mode so untested)
12.2.1.3
updateDomain in online mode with domain_type RestrictedJRF
12.2.1.3
12.2.1.4
updateDomain in online mode with domain_type JRF
(open issue on 10.3.6 in online mode, so untested)
12.2.1.3
12.2.1.4


core/src/main/python/wlsdeploy/tool/create/domain_typedef.py
```diff
@@ -109,6 +109,13 @@ def domain_type_has_jrf_resources(self):
         """
         return 'JRF-MAN-SVR' in self.get_server_groups_to_target()
 
+    def domain_type_is_restricted_jrf(self):
+        """
+        A workaround for a bug with applyJRF in online with RestrictedJRF
+        :return:
+        """
+        return self.get_domain_type() == 'RestrictedJRF'
+
```
rakillen: CER and reviewers - do you think there is a case where a user might create their own typedef based on RestrctedJRF?  We could add a flag instead of checking the name directly, but not sure if it is worth it.
I noticed there is another direct name check in this file for JRF, but it is not used any more.

CarolynRountree: This is a good point. I put it in as a temporary until come up with a replacement. But once in the code, it probably will not come out. I do not know how many upper stack products use RestrictedJRF instead of JRF when installing their own product templates. I don't think we can use the server groups for RestrictedJRF, but we could look at the templates. This is a bit of code because early version have the physical name of the template jar and later versions, have the template name (from the template-info.xml)

CarolynRountree:
Made changes as requested for check, messages and method names. Merged master and did spot retesting.

PR #375, opened 2019-05-09
Wdt 334 attributes not in lsa map not discovered

CarolynRountree:
This PR must go into a release at the same time as WDT-338


core/src/main/python/wlsdeploy/tool/util/mbean_utils.py
```diff
@@ -0,0 +1,767 @@
...
+    Utility class used to provide information about WLST attributes as retrieved from the MBean MBeanInfo or Interface
+    methods. This class has methods to provide the information stored in different combinations. All methods that want
+    to combine the information from the MBean helpers into different combinations are located in this class.
+    """
+
+    def __init__(self, model_context, exception_type):
+        self.__model_context = model_context
+        self.__exception_type = exception_type
+        self.__aliases = Aliases(self.__model_context, wlst_mode=self.__model_context.get_target_wlst_mode())
+        self.__alias_helper = AliasHelper(self.__aliases, _logger, exception_type)
```
rakillen: I think we should not construct the Aliases object, but pass it in.  It takes some effort to re-build it per WLS version.  Maybe just pass AliasHelper around, it should have all the methods, and the exception type is baked in.

core/src/main/python/wlsdeploy/tool/util/mbean_utils.py
```diff
@@ -0,0 +1,767 @@
...
+    specific information.
+    """
+
+    __interface_matcher = re.compile('Bean$')
+
+    def __init__(self, model_context, exception_type, location):
+        self.__model_context = model_context
+        self.__exception_type = exception_type
+        self.__location = location
+        self.__aliases = Aliases(self.__model_context, wlst_mode=self.__model_context.get_target_wlst_mode())
```
rakillen: I think we should not construct the Aliases object, but pass it in.  It takes some effort to re-build it per WLS version.  Maybe just pass AliasHelper around, it should have all the methods, and the exception type is baked in.

CarolynRountree: I would be nice to pull that exception type OUT of the model_context - would be useful for any place that would like to generate and throw a tool level  exception. It's being passed around to those classes that do not generate their own exception types.

Speaking of which, I considered a separate type for this class - but the class is just so small and single use, I decided not to. Opinions?

PR #391, opened 2019-06-06
Addbootproperties

jshum2479:
add boot.properties to the security folder at the end of domain creation

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/create/domain_creator.py
+++ b/core/src/main/python/wlsdeploy/tool/create/domain_creator.py
@@ -1099,6 +1099,8 @@ class DomainCreator(Creator):
         return
 
     def __create_boot_dot_properties(self):
+        _method_name = '__create_boot_dot_properties'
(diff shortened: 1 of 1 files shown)


core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -161,6 +165,7 @@ def create(self):
         self.__fail_mt_1221_domain_creation()
         self.__create_domain()
         self.__deploy()
+        self.__create_boot_dot_properties()
```
ddsharpe: This should be optional.  The default WebLogic behavior is to never do boot.properties for production mode.  We should not make this automatic.  But, I think its a good feature to have that is especially useful for the Docker/K8S use case.  Just keep in mind that Docker/K8S is not the ONLY use case for WDT.

ddsharpe:
This needs to be corrected per my above comment, or reverted.

PR #392, opened 2019-06-06
Addopssencryptkey

jshum2479:


core/src/main/python/wlsdeploy/util/wlst_helper.py
```diff
@@ -1395,3 +1395,16 @@ def reopen_offline(domain_home):
...
     _logger.exiting(class_name=_class_name, method_name=_method_name)
+
+
+def set_shared_secret_store_with_password(wallet_path, password):
+    """
+    Set opss store password
+    :param wallet_path: opss extracted wallet
+    :param password:  opss store extraction time password
+    """
+    _method_name = 'setSharedSecretStoreWithPassword'
```
rakillen: doesn't match method name set_shared_secret_store_with_password

core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -1128,3 +1134,51 @@ def __create_boot_dot_properties(self):
             ostream.close()
         self.logger.exiting(class_name=self.__class_name, method_name=_method_name)
         return
+
+    def __configure_opss_secrets(self):
```
ddsharpe: Dont we need to screen off the online use case?  You cannot update the secret in online mode, can you?

jshum2479: It's for offline , not online

jshum2479: and during creation.

jshum2479: i.e. it's not even in the code path for online

jshum2479:
@CarolynRountree   @ddsharpe 

1.  Should this be documented and where
2.  I am trying a place to move the common python code.
3.  There is no discovery for now.

PR #396, opened 2019-06-11
Jira:WDT-346 fix reported issue and other known problems with dynamic…

CarolynRountree:
… cluster targeting


core/src/main/python/wlsdeploy/tool/util/target_helper.py
```diff
@@ -4,7 +4,9 @@
 """
 
 import copy
-import oracle.weblogic.deploy.util.StringUtils as StringUtils
+
+import java.lang.Boolean as Boolean
```
rakillen: I don't think this import is used?

KnownIssues.md
```diff
@@ -1,5 +1,18 @@
 The following list are known issues. The issue may contain a work-around or an associated Issue number.
 
+**ISSUE**:
+   The createDomain tool does not target non-JRF product resources to dynamic clusters.
+   WebLogic WLST will not assign resources associated with extension template or custom template user server groups
+   to dynamic clusters or dynamic servers. The createDomain can target any resources associated to JRF
+   related user server groups using the FMW WLST extension function applyJRF.
+
+**ACTION**:
```
ddsharpe: This section is unclear.  Are you asking the user to do all three of these tasks, and if so, how does each step resolve their issue.  The issue section needs more description of what the user sees that triggers the search that led them here to "known issues".  For example, if I see "A", then this issue applies to you.

PR #407, opened 2019-06-29
run decrypt for admin credentials before creating boot.properites as …

jshum2479:
decrypt admin credentials from model before creating boot.properties


core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -1121,6 +1120,9 @@ def __create_boot_dot_properties(self):
 
         for server in servers:
             properties = Properties()
+
+            admin_username = self.aliases.decrypt_password(admin_username)
```
ddsharpe: Why do you need to decrypt it each time there is a server (for every server)?  Shouldn't this decrypt step be done near line 1110 (outside of the for-loop)?

jshum2479: Yea, I know about that.  Both decrypt and encrypt of credentials can be outside the loop, I also change to create boot.properties when it is not in production mode

core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -167,6 +169,7 @@ def create(self):
         self.__fail_mt_1221_domain_creation()
         self.__create_domain()
         self.__deploy()
+        self.__create_boot_dot_properties()
```
ddsharpe: Why was this added back?  Shouldn't this be removed?

ddsharpe: verbally discussed.  boot.properties is only created when startup is set to development mode (not in production mode).  This is the same behavior that the WLS config tool uses.

PR #411, opened 2019-07-19
WDT-353: Add WLSRoles section to domainInfo

TheFrogPad:
Add WDT model support to specify WebLogic global roles and add documentation around existing `Security` section that specifies users and groups. Additional details on `WLSRoles` section can also be found in the JIRA task WDT-353.


TheFrogPad:
Hmmm... I have no clue why the last merge on the branch would fail wercker build as this was to update a simple typo in the markdown file plus I have no permission on the wercker project to see what may be wrong?

core/src/main/python/wlsdeploy/tool/validate/validator.py
```diff
@@ -387,7 +400,8 @@ def __validate_domain_info_section(self, model_section_key, model_dict, validati
             if '${' in section_dict_key:
                 validation_result.add_error('WLSDPLY-05035', model_folder_path, section_dict_key)
 
-            self._logger.finer('WLSDPLY-05011', section_dict_key, section_dict_value,
```
rakillen: good catch here - yikes!

site/security_users_groups_roles.md
```diff
@@ -0,0 +1,53 @@
+### Modeling WebLogic Users, Groups and Roles
+WebLogic Server has the ability to establish a set of Users, Groups and Global Roles as part of the WebLogic domain creation. The WebLogic global roles become part of the WebLogic role mapper (i.e. `XACMLRoleMapper`) and are specified under `domainInfo` in the `WLSRoles` section. The users and groups become part of the Embedded LDAP server (i.e. `DefaultAuthenticator`) and are specified under `topology` in the `Security` section.
```
rosemarymarano: Users, Groups and Global Roles -> users, groups, and global roles (these terms do not need to be capitalized)

site/security_users_groups_roles.md
```diff
@@ -0,0 +1,53 @@
...
+
+#### WebLogic Global Roles
+The model allows for the definition of WebLogic roles that can augment the well known WebLogic global roles (e.g. `Admin`, `Deployer`, `Monitor`, ...) in addition to defining new roles. When updating the well known WebLogic roles an `UpdateMode` can be specified as `{ append | prepend | replace }` with the default being `replace` when not specified. Also, when updating the well known roles, the specified `Expression` will be a logical `OR` with the default expression. The `Expression` value for the role is the same as when using the WebLogic `RoleEditorMBean` for a WebLogic security role mapping provider.
```
rosemarymarano: When updating the well known WebLogic roles an -> When updating the well known WebLogic roles, (comma) an

site/security_users_groups_roles.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+The `Admin` role will have the expression appended to the default expression, the `Deployer` role expression will replace the default, the `Monitor` role expression will be prepended to the default expression and `Tester` will be a new role with the specified expression.
+
+In addition, the `Expression` value can use the variable placeholder syntax specified when running the [Create Tool](create.md) as shown in the above example.
+
+#### WebLogic Users and Groups
+The model allows for the definition of a set of users and groups that will be loaded into the WebLogic Embedded LDAP Server (i.e. `DefaultAuthenticator`). New groups can be specified and users can be added as members of the new groups or existing groups such as the `Administrators` group which is defaulted to be in the WebLogic `Admin` global role.
+
+The user password can be specified with a placeholder or encrypted with [Encrypt Tool](encrypt.md). An example `Security` section that adds an additional group `AppMonitors`, adds two new users and places the users into groups is as follows:
```
rosemarymarano: encrypted with [Encrypt Tool]-> encrypted with the [Encrypt Tool]

site/security_users_groups_roles.md
```diff
@@ -0,0 +1,53 @@
+### Modeling WebLogic Users, Groups and Roles
```
rosemarymarano: WebLogic Users, Groups and Roles -> WebLogic Users, Groups, (comma) and Roles (our style guide dictates the use of the Oxford (serial) comma: in a list of three or more items, separate the items with commas, including a comma before the conjunction. Please make this correction globally.

TheFrogPad: This and the other comments have been addressed.

site/security_users_groups_roles.md
```diff
@@ -51,3 +51,4 @@ topology:
 - The processing of users, groups, and roles will only take place when using the [Create Domain Tool](create.md)
 - WebLogic global roles are only supported with WebLogic Server version 12.2.1 or greater
 - WebLogic global roles are only updated for the WebLogic security XACML role mapping provider (i.e. `XACMLRoleMapper`)
+- The user and group processing is not complete, groups are created but users are only assigned to the `Administrators` group
```
ddsharpe: Please change this line to include "due to a bug in WLST".  Something like, "Due to a bug in WLST, Users cannot be assigned to groups.  Users created this way are automatically added to the Administrators group and are not added to the group(s) specified. As soon as a patch is available for WLST, we will post it here."

rosemarymarano: As far as I know, we've never used the wording, "Due to a bug, ..." If it's important to you to state that, ok. Or, an alternative wording to consider is, " Currently, users cannot be assigned to groups. Users created this way (using WLST?) are automatically added to the Administrators group and are not added to the group(s) specified. As soon as a patch to correct this behavior is available, we will post it here."

TheFrogPad: I have made a update to the wording based on Rosemary's suggested text so I think we should be good now. Thanks.

PR #414, opened 2019-08-12
Extractdiffloc

jshum2479:
I want to add a flag -store_model <1>  to the WDT.  If the argument -store_model 1 is passed for createDomain,  the merged model is stored in the $domain/wlsdeploy/domain_saved.json - a fixed location.   This is only used by the operator temporarily during introspect job.


jshum2479:
I changed to use environment variable __STORE_MODEL__ instead

core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -345,6 +345,12 @@ def __create_domain(self):
         self.library_helper.install_domain_libraries()
         self.library_helper.extract_classpath_libraries()
         self.wlsroles_helper.process_roles()
+        if os.environ.has_key('__STORE_MODEL__'):
```
rakillen: Any interest in changing this to "WLSDEPLOY_STORE_MODEL"?  They do this elsewhere in the code, such as WLSDeployLoggingConfig.java .  I could live with either, though

PR #416, opened 2019-08-13
Correct WLS build version to 12.2.1.3

ddsharpe:
The build should be using WLS 12.2.1.3 to run unit tests.


ddsharpe:
This change is superseded by the move off of Quay

PR #422, opened 2019-08-19
Issue 144 remove named objects

### LINKED ISSUE #144: Enhancement request: Add removal of objects into the MetaData model
From the README explanation it is clear that renaming existing objects (like Managed Servers) is not supported. 
> ...the WebLogic Deploy Tooling does not support the use of the Name attribute in these folders and any attempt to set the Name attribute will not be honored.

But would it be possible to model a "not wanted object any more" into the model? Which will be deleted when the domain is `updateDomain` (or even on `createDomain` where the template adds a not wanted object)?
Possible modelling solution in yaml `model_file` could be to put an exclamation mark "!" in front of the object name, causing it to be deleted or removed to be set. 

(Rationale: When creating a default Forms domain, the managed server from the extentionTemplate "Oracle Forms" is defaulted to "WLS_FORMS" and cannot be overwritten with the `createDomain` resulting in a obsolete Managed Server after initial run.

rakillen:
Fixes #144.


core/src/main/python/wlsdeploy/tool/deploy/deployer_utils.py
```diff
@@ -177,6 +177,49 @@ def merge_lists(existing_list, model_list, separator=',', return_as_string=True)
...
+    :param name: the name to be checked
+    :return: True if the name is prefixed, false otherwise
+    """
+    return name.startswith("!")
+
+
+def get_delete_item_name(name):
+    """
+    Returns the WLST name of the item to be deleted.
+    Removes the "!" prefix from the name. Assumes the name has been checked with is_delete_name().
```
ddsharpe: The doc says it "assumes". Looking at the code, that does not appear to be true since it actually checks "is_delete_name()"

core/src/main/python/wlsdeploy/tool/deploy/deployer_utils.py
```diff
@@ -177,6 +177,49 @@ def merge_lists(existing_list, model_list, separator=',', return_as_string=True)
...
+    :param name: the prefixed model name of the item to be deleted
+    :return: the model name of the item to be deleted
+    """
+    if is_delete_name(name):
+        return name[1:]
+    return name
```
ddsharpe: Seems like this should return null or throw an exception.  Returning the name does not feel right.

core/src/main/python/wlsdeploy/tool/deploy/deployer_utils.py
rakillen: Fixed comment to match changes

core/src/main/python/wlsdeploy/tool/deploy/deployer_utils.py
rakillen: Fixed to throw exception

README.md
```diff
@@ -220,6 +221,36 @@ In the example above, the `Target` attribute is specified three different ways,
...
 
+### Declaring Named MBeans to Delete
+
+With WebLogic Deploy Tooling release 1.3.0, you can specify named items in the model to be deleted using the Create Domain, Update Domain, Deploy Applications Tools.  Named items are those that have multiple instances that are distinguished by user-provided names, such as managed servers, datasources, and security realms.  Items to be deleted are prepended with an exclamation point (!) in the model.
```
rosemarymarano: Update Domain, Deploy Applications Tools. -> Update Domain, and Deploy Applications Tools.
datasources -> data sources ((adj) (n) Two words.)

rakillen: Changed these + one more data source

README.md
```diff
@@ -220,6 +221,36 @@ In the example above, the `Target` attribute is specified three different ways,
...
+            newrealm:
+                AuthenticationProvider:
+                ...
+```
+
+This feature does not apply to named security providers within a realm. These items follow a special set of rules that are required to maintain their ordering. See [Modeling Security Providers](site/security_providers.md) for detailed information.
+
+This feature cannot be use to un-deploy applications or remove libraries.
+
+This feature can also be used for simple updates to domain configuration. In this example, the managed server ```obsoleteServer``` will be deleted, and ```newServer``` will be created:
```
CarolynRountree: This seems like a dangling paragraph. You already discussed the purpose in the first paragraph. Now I scratch my head and ask what does he mean here? Is he saying that making a simple update to a server would be to remove it and re-add it? You said something different in the other paragraph

core/src/main/python/wlsdeploy/tool/deploy/jms_resources_deployer.py
```diff
@@ -151,10 +154,16 @@ def _add_jndi_properties(self, property_name_nodes, location):
         name_attribute = self.alias_helper.get_wlst_attribute_name(properties_location, KEY)
         mbean_type = self.alias_helper.get_wlst_mbean_type(properties_location)
 
+        for property_name in property_name_nodes:
+            if deployer_utils.is_delete_name(property_name):
+                name = deployer_utils.get_delete_item_name(property_name)
```
CarolynRountree: Just curious. Is WLST nice enough to remove the deleted server from all targets that reference it?

rakillen: I tried this by removing a server that is referenced as a target.  Offline, I got a pretty good error about wlst.set(Target, m2).  Offline, it has a better message, because we attempt to look up the server and don't find it.

README.md
rakillen: Re-ordered this with simple example first, then template/security stuff

core/src/main/python/wlsdeploy/tool/deploy/jms_resources_deployer.py
CarolynRountree: Did you see the comment at top of file? Related to this.

PR #425, opened 2019-08-21
Wdt# support for 12214

CarolynRountree:
Added JMS load balancing fields for 12.2.1.4. Cleaned up typedef files for 12.2.1.4


core/src/main/typedefs/WLS.json
```diff
@@ -9,7 +9,8 @@
         "12.1.2": "WLS_12CR1",
         "12.1.3": "WLS_12CR1",
         "12.2.1": "WLS_12CR2",
-        "19.1":   "WLS_19CR1"
+		"12.2.1.3": "WLS_12CR2",
```
rakillen: line 12 has 2 tabs instead of spaces

CarolynRountree:
I added the changes to alias_constants and aliases.py for my aliases definition tests. I used these tests to gather information about the new attribute fields.

core/src/main/python/wlsdeploy/aliases/aliases.py
```diff
@@ -1130,6 +1131,39 @@ def get_ignore_attribute_names(self):
...
+        result = None
+        attribute_info = self._alias_entries.get_alias_attribute_entry_by_model_name(location, model_attribute_name)
+        if attribute_info is not None and PREFERRED_MODEL_TYPE in attribute_info:
+            result = attribute_info[PREFERRED_MODEL_TYPE]
+        self._logger.exiting(class_name=self._class_name, method_name=_method_name, result=result)
+        return result
+
+    def get_wlst_read_type(self, location, model_attribute_name):
+        """
+        Return the wlst read type, which overrides the wlst type when retrieving the attribute value.
```
rakillen: capitalize WLST in comments

core/src/main/typedefs/RestrictedJRF.json
```diff
@@ -5,8 +5,8 @@
     "description": "Restricted JRF type domain definitions",
     "versions": {
         "12.2.1": "RJRF_12CR2",
-        "12.2.1.3": "RJRF_12213",
-        "19.1":   "RJRF_19CR1"
+		"12.2.1.3": "RJRF_12CR2",
+		"12.2.1.4": "RJRF_12CR2"
```
rakillen: lines 8, 9 have tabs instead of spaces

CarolynRountree:
It is not necessary to put all versions in the 'versions' section of the typedef files. However, the files are inconsistent - If you look at WLS, 12.1.1 and 10.3.6 both use 11G. 12.2.1 has 12C, but versions 12.2.1.1 and 12.2.1.2 are not in the versions section. I do not have to put 12.2.1.3 and 12.2.1.4, I added them for clarity that those releases are supported. I suppose we could just update with the last release, or just not put them in at all until needed. For instance, 12.3 (I believe) will require an entry. I know an entry is required when the first number changes, and I believe it is needed for changes to the second number also

core/src/main/typedefs/RestrictedJRF.json
CarolynRountree: changed indents to spaces and pushed

core/src/main/python/wlsdeploy/aliases/aliases.py
```diff
@@ -1130,6 +1131,40 @@ def get_ignore_attribute_names(self):
         """
         return self._alias_entries.IGNORE_FOR_MODEL_LIST
 
+    def get_preferred_model_type(self, location, model_attribute_name):
+        """
+        Return the preferred model type, if present, for the alias attribue definition
```
rakillen: *attribute

PR #440, opened 2019-09-11
Opsscli

jshum2479:
Add opss wallet import command line options for sharing DB use case

-opss_wallet <wallet directory>
-opss_wallet_passphrase <passphrase>

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/create/domain_creator.py
+++ b/core/src/main/python/wlsdeploy/tool/create/domain_creator.py
@@ -1175,5 +1175,11 @@ class DomainCreator(Creator):
                     zip_entry = archive_file.getOPSSWallet();
                     FileUtils.extractZipFileContent(archive_file, zip_entry, extract_path)
                     self.wlst_helper.setSharedSecretStoreWithPassword(extract_path, opss_secret_password)
-        self.logger.exiting(class_name=self.__class_name, method_name=_method_name)
-        return extract_path
+            else:
+                opss_secret_password = self.model_context.get_opss_wallet_passphrase()
+                opss_wallet = self.model_context.get_opss_wallet()
+                if opss_wallet is not None and opss_secret_password is not None:
+                    self.wlst_helper.setSharedSecretStoreWithPassword(opss_wallet, opss_secret_password)
+
+            self.logger.exiting(class_name=self.__class_name, method_name=_method_name)
+        return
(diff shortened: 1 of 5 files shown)


core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -1175,5 +1175,11 @@ def __configure_opss_secrets(self):
...
                     self.wlst_helper.setSharedSecretStoreWithPassword(extract_path, opss_secret_password)
-        self.logger.exiting(class_name=self.__class_name, method_name=_method_name)
-        return extract_path
+            else:
+                opss_secret_password = self.model_context.get_opss_wallet_passphrase()
+                opss_wallet = self.model_context.get_opss_wallet()
+                if opss_wallet is not None and opss_secret_password is not None:
+                    self.wlst_helper.setSharedSecretStoreWithPassword(opss_wallet, opss_secret_password)
+
+            self.logger.exiting(class_name=self.__class_name, method_name=_method_name)
```
rakillen: Exiting log should be same indent level as return, I think?

core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -1175,5 +1175,11 @@ def __configure_opss_secrets(self):
                     zip_entry = archive_file.getOPSSWallet();
                     FileUtils.extractZipFileContent(archive_file, zip_entry, extract_path)
                     self.wlst_helper.setSharedSecretStoreWithPassword(extract_path, opss_secret_password)
-        self.logger.exiting(class_name=self.__class_name, method_name=_method_name)
-        return extract_path
+            else:
```
rakillen: Hard to tell from this diff, but it looks like this "else" is for "if OPSS_SECRETS in domain_info", still inside "if domain_info is not None".  Seems like it should be:
```
if (domain_info is not None) and (OPSS_SECRETS in domain_info):
    ... use values from domain info
else:
    ... check model context
```
Even though the domain_info section should always be present on a create.

core/src/main/resources/oracle/weblogic/deploy/messages/wlsdeploy_rb.properties
```diff
@@ -1442,6 +1442,8 @@ WLSDPLY-20023={0} unable to add model file {1} to archive as {2}: {3}
 WLSDPLY-20024={0} failed to persist the model to the archive file {1}: {2}
 WLSDPLY-20025=For {0}, specify the {1} or {2} argument, but not both
 WLSDPLY-20026={0} failed to find a model file in archive {1}, and {2} argument not specified
+WLSDPLY-20027=Enter the opss wallet passphrase
+WLSDPLY-20028=Failed to read the opss wallet passphrase input from the user: {0}
```
rakillen: OPSS should be all caps

core/src/main/python/wlsdeploy/tool/create/domain_creator.py
jshum2479: Yes.  I have refactored the code and corrected the spelling in message catalog

PR #441, opened 2019-09-12
Change standalone validator to warn FILE variable substitution instead of failure

jshum2479:
Change standalone validator so that if the variable @@FILE@@ value does not exist, it warns instead of error out.  The environment where it is validating may not be have the file contents yet.


core/src/main/python/wlsdeploy/util/variables.py
```diff
@@ -243,9 +246,18 @@ def _read_value_from_file(file_path):
...
-        raise ex
+        if model_context.get_validation_method() == 'strict':
+            if validation_result:
+                validation_result.add_error('WLSDPLY-01733', file_path, e.getLocalizedMessage())
+            ex = exception_helper.create_variable_exception('WLSDPLY-01733', file_path, e.getLocalizedMessage(), error=e)
+            _logger.throwing(ex, class_name=_class_name, method_name=method_name)
+            raise ex
+        else:
+            if validation_result:
+                validation_result.add_warning('WLSDPLY-01733', file_path, e.getLocalizedMessage())
```
tbarnes-us: Since missing files are fully expected, can this be an Info instead of a Warning?  Warnings can be distracting as they imply there's something that should be investigated...

jshum2479: No, the tool is not only used by the k8s and imagetool.  It can be used on premise and other situation

tbarnes-us: Then add a third state or check for env var?  It's distracting to get spurious warnings.

rakillen: There's a new PR for this change, and a discussion of this topic.
https://github.com/oracle/weblogic-deploy-tooling/pull/443

PR #464, opened 2019-10-18
Wdt support for 1411

CarolynRountree:
Add support for WebLogic 14.1.1
1. Add version for domain typedefs. Put not supported in JRF and RestrictedJRF
2. Changes to python code to support jython 2.7.1 runtime issues
   a. Strings from Class different
   b. fix for module main
   c. unicode fixes
   d. Change PyOrderedDictionary to not reference PyDictionary map directly, and compile with 2.2.1 only.
 3. pom changes for future
4. changes for 2.a. extensive in mbean_utils


CarolynRountree:
Yes the unicode is supported in 2.2.1 python. Jython did not handle this. We continue to compile java code with jython 2.2.1 (for PyOrderedDict, for example) and run under 2.7.1. This has been tested in 14.1.1, 12.2.1.4 and 12.2.1.3

PR #466, opened 2019-10-23
Wdt#457 capture user override scripts

### LINKED ISSUE #457: Discover domain should capture setUserOverrides domain shell script in archive
In setDomainEnv.sh, script looks for setUserOverrides.sh in the domain to apply domain env customizations.  This should be captured by discovery so that it can be replayed by create.

See excerpt from setDomainEnv.sh:
# Set user overrides, if available.

if [ -f ${DOMAIN_HOME}/bin/setUserOverrides.sh ] ; then
	. ${DOMAIN_HOME}/bin/setUserOverrides.sh
fi

CarolynRountree:
Collect user overrides environment scripts from domain/bin into the archive. Install the scripts during create/update into the target domain/bin, or copy from the designated location. Only the scripts that are identified in the domainInfo: domainBin section will be extracted/copied into the target domain. Fixes #457


CarolynRountree:
[user_overrides.zip](https://github.com/oracle/weblogic-deploy-tooling/files/3763519/user_overrides.zip)

Attached are the results from my local test.

PR #467, opened 2019-10-23
Fix delete notation merge where no variables are used

ddsharpe:
Delete notation was not merged correctly when variables are not used.  Also, changed behavior slightly so that delete notation is not saved when finding a match in a previous model (deletes from model, but not saved in final model).


rakillen:
Good catch on variables, d’oh!

 

The second change breaks the following case for update, I think:

 

-          Domain has server m2

-          First model has m2

-          Second model has !m2

 

The merged model has no m2 or !m2 entry, so m2 remains in the domain, when the intent was to delete it.

 

Richard

PR #468, opened 2019-10-23
fix for domain discover issue found in 10.3.6 but not 12.2.1.3

CarolynRountree:
14.1.1 version of jython does not create strings for instances the same as in the early version of jython. The change made to handle all versions worked for 14.1.1.0.0 and 12.2.1.3.0 but broke in 10.3.6. Made a fix to handle for all versions.

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/discover/discoverer.py
+++ b/core/src/main/python/wlsdeploy/tool/discover/discoverer.py
@@ -4,8 +4,6 @@ Licensed under the Universal Permissive License v 1.0 as shown at https://oss.or
 """
 import os
 
-import java.lang.Exception as JException
-
 from oracle.weblogic.deploy.aliases import AliasException
 from oracle.weblogic.deploy.discover import DiscoverException
 from oracle.weblogic.deploy.util import PyOrderedDict as OrderedDict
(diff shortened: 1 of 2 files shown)


core/src/main/python/wlsdeploy/tool/discover/discoverer.py
```diff
@@ -727,6 +729,15 @@ def _massage_online_folders(lsc_folders):
     return folder_list
 
 
+def get_interface_name(mbean_interface):
```
ddsharpe: Why does this method exist in two places?  mbean_utils and discoverer

CarolynRountree: This will take me some time to combine the two together. It is not a straight up conversion. Putting on the back burner for now.

core/src/main/python/wlsdeploy/tool/discover/discoverer.py
```diff
@@ -4,6 +4,8 @@
 """
 import os
 
+import java.lang.Exception as JException
```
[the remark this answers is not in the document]
CarolynRountree: The get_interface_name used it, so now it is no longer used.

PR #470, opened 2019-10-28
Add the domain name and domainInfo admin credentials automatically in…

CarolynRountree:
The discover tool will now add the admin credentials and the domain name to the persisted model. Have attached an example.
[domain-info.txt](https://github.com/oracle/weblogic-deploy-tooling/files/3779686/domain-info.txt)

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/discover/domain_info_discoverer.py
+++ b/core/src/main/python/wlsdeploy/tool/discover/domain_info_discoverer.py
@@ -10,6 +10,7 @@ from java.io import File
 from oracle.weblogic.deploy.util import WLSDeployArchiveIOException
 from oracle.weblogic.deploy.util import FileUtils
 
+from wlsdeploy.aliases import alias_constants
 from wlsdeploy.aliases import model_constants
 from wlsdeploy.aliases.wlst_modes import WlstModes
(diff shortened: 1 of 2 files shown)


core/src/main/python/wlsdeploy/tool/discover/domain_info_discoverer.py
```diff
@@ -40,13 +41,18 @@ def discover(self):
...
+        self.add_admin_credentials()
         model_top_folder_name, result = self.get_domain_libs()
         discoverer.add_to_model_if_not_empty(self._dictionary, model_top_folder_name, result)
         model_top_folder_name, result = self.get_user_env_scripts()
         discoverer.add_to_model_if_not_empty(self._dictionary, model_top_folder_name, result)
         _logger.exiting(class_name=_class_name, method_name=_method_name)
         return self._dictionary
 
+    def add_admin_credentials(self):
+        self._dictionary[model_constants.ADMIN_USERNAME] = alias_constants.PASSWORD_TOKEN
```
ddsharpe: Can the Admin username be discovered?

CarolynRountree: No. It is a property field that becomes hidden after the writeDomain is performed.

PR #528, opened 2020-01-23
remove wlstModule usage and compress 3 wlst helper classes into 1

CarolynRountree:
Use the wlst globals for calling WLST commands instead of the wlst.module. This will make available commands not in the WLSTModule. These commands will all use the same WLST context, which means for updates like apply_jrf, not required to update the domain and then open a separate context.

The two different wlst_helper.py have been combined into a single file. I suggest you look at the source file instead of the diffs as effectively the tool/utils/wlst_helper.py was deleted and the utils/wlst_helper.py moved to tool/utils/wlst_helper.py.

I would like to get this merged into master so we can work with the changes for some time before the next release.

Do not close the Issue, as another round to address the many updates required by using applyJRF command will be removed to address the problem in this Issue


core/src/main/python/wlsdeploy/tool/util/wlst_helper.py
```diff
@@ -1,228 +1,264 @@
...
+import oracle.weblogic.deploy.util.StringUtils as StringUtils
+import weblogic.management.mbeanservers.edit.ValidationException as ValidationException
 
 from wlsdeploy.exception import exception_helper
-from wlsdeploy.util import wlst_extended
-from wlsdeploy.util import wlst_helper
+from wlsdeploy.logging.platform_logger import PlatformLogger
+
+wlst_functions = None
+wlst_state = None
```
rakillen: what's wlst_state?  couldn't find references to it

core/src/main/python/wlsdeploy/tool/util/wlst_helper.py
```diff
@@ -1,228 +1,264 @@
...
         return result
 
-    def set(self, attribute_name, attribute_value, masked=False):
+    def set_if_needed(self, wlst_name, wlst_value, masked=False):
+        """
+        Set the WLST attribute to the specified value if the name and value are not None.
+        :param wlst_name: the WLST attribute name
+        :param wlst_value: the WLST attribute value
+        :param masked: whether or not to mask the value in the logs, default value is False
+        :raises: CreateException: if an error occurs
```
rakillen: tool Exception

core/src/main/python/wlsdeploy/tool/util/wlst_helper.py
```diff
@@ -236,852 +272,1553 @@ def create_and_cd(self, alias_helper, type_name, name, location, create_path=Non
         :raises: BundleAwareException of the specified type: if an error occurs
```
rakillen: tool Exception (just to be consistent)

core/src/main/python/wlsdeploy/tool/util/wlst_helper.py
```diff
@@ -236,852 +272,1553 @@ def create_and_cd(self, alias_helper, type_name, name, location, create_path=Non
...
+        self.__logger.exiting(class_name=self.__class_name, method_name=_method_name)
 
     def set_option_if_needed(self, option_name, option_value):
         """
         Set the WLST domain option to the provided value if the name and value are not None.
 
         :param option_name: attribute name at the current location
         :param option_value: to configure the attribute
+
         :raises: BundleAwareException of the specified type: if an error occurs
```
rakillen: tool Exception (just to be consistent)

core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -959,8 +958,7 @@ def __configure_fmw_infra_database(self):
             self.wlst_helper.cd(wlst_path)
             wlst_name, wlst_value = \
                 self.alias_helper.get_wlst_attribute_name_and_value(location, DRIVER_PARAMS_PROPERTY_VALUE, stb_user)
-            self.wlst_helper.set_if_needed(wlst_name, wlst_value,
-                                           JDBC_DRIVER_PARAMS_PROPERTIES, DRIVER_PARAMS_USER_PROPERTY)
+            self.wlst_helper.set_if_needed(wlst_name, wlst_value,)
```
rakillen: extra comma at the end

PR #550, opened 2020-02-21
Use Jython directly for encryptModel

rakillen:
JIRA WDT-387, part 1

Created shared functions for Jython tools.
Refactor validateModel to use shared functions.

extractDomainResource and injectVariables will be in a separate PR


installer/src/main/bin/encryptModel.cmd
```diff
@@ -21,298 +21,64 @@
...
-@rem     - -domain_type        The type of domain to create. This argument is
-@rem                           is optional.  If not specified, it defaults to WLS.
-@rem
-@rem     - -wlst_path          The path to the Oracle Home product directory under
-@rem                           which to find the wlst.cmd script.  This is only
-@rem                           needed for pre-12.2.1 upper stack products like SOA.
-@rem
-@rem                           For example, for SOA 12.1.3, -wlst_path should be
-@rem                           specified as %ORACLE_HOME%\soa
+@rem     -oracle_home
```
rakillen: CER has suggested getting rid of the "The script uses... directly" section in these files.  I agree, it doesn't add any value, and adds confusion about what the full set of arguments includes.  Opinions?

installer/src/main/bin/shared.cmd
```diff
@@ -0,0 +1,250 @@
...
+    )
+
+    SET JVM_SUPPORTED=1
+    IF %JVM_VERSION_PART_ONE% LEQ 1 (
+        IF %JVM_VERSION_PART_TWO% LSS %MIN_JDK_VERSION% (
+            SET JVM_SUPPORTED=0
+            IF %JVM_VERSION_PART_TWO% LSS 7 (
+                ECHO You are using an unsupported JDK version %JVM_FULL_VERSION% >&2
+            ) ELSE (
+                ECHO JDK version 1.8 or higher is required when using encryption >&2
```
CarolynRountree: Isn't the encrypt jdk check already in the encrypt model? Is this print necessary?

installer/src/main/bin/shared.cmd
```diff
@@ -0,0 +1,250 @@
...
+    SET PY_SCRIPTS_PATH=%WLSDEPLOY_HOME%\lib\python
+
+    ECHO ^
+    %JAVA_HOME%/bin/java -cp %CLASSPATH% ^
+        %JAVA_PROPERTIES% ^
+        org.python.util.jython ^
+        "%PY_SCRIPTS_PATH%\%JYTHON_SCRIPT%" %SCRIPT_ARGS%
+
+    %JAVA_HOME%/bin/java -cp %CLASSPATH% ^
+        %JAVA_PROPERTIES% ^
```
CarolynRountree: The WLSDEPLOY_PROPERTIES is usually used to set WLST specific properties. I have always had a problem with the name of this property because I cannot think of a property that is specific to WDT that is set using this value. Unless its the typedef json file which isn't used in these scripts anyway. Opinion?

installer/src/main/bin/encryptModel.cmd
CarolynRountree: If we put anything up here it should only be the environment variables specific to the script. These aren't described in the usage statement.

installer/src/main/bin/shared.cmd
jshum2479: WLSDEPLOY_PROPERTIES can be used by customer.  For example, setting -Doracle.jdbc.fanEnabled=false when creating a JRF domain against RAC.  WLST_PROPERTIES is reset in the script.  If we remove this, we need to document the changes but may break existing customer if they are using it.

CarolynRountree: This shared script is for non-WLST scripts only, correct?

installer/src/main/bin/shared.cmd
rakillen: encryptModel just passes in 8 as the required version, and this function logs and returns the exit code if the version is lower.

installer/src/main/bin/shared.cmd
rakillen: This will eventually be the one for WLST scripts also.  It will re-use some of these methods.

CarolynRountree: so back to the original comment. Do we need this property if we have JAVA_PROPERTIES? in the other scripts we set the WLST command argument to this value. In the encryptModel script (or the others that will use this shared script), you will never have the setting -Doracle.jdbc.fanEnabled=false

ddsharpe: To change the name of the property would break backward compatibility.  There needs to be a really REALLY good reason to do so.

CarolynRountree: There is no backward compatibility. WLSDEPLOY_PROPERTIES was set on WLST_PROPERTIES (or whatever the wlst environment var is) and wlst does not exist in this new script.

PR #556, opened 2020-02-26
Clarify RCU message, reduce level to FINER

rakillen:
JIRA-WDT-389

This scenario occurs when a domain type with extension/custom templates is created, and is not JRF (currently only RestrictedJRF).


core/src/main/resources/oracle/weblogic/deploy/messages/wlsdeploy_rb.properties
```diff
@@ -1168,7 +1168,7 @@ WLSDPLY-12246=Adding custom extension template file {0} to domain
 WLSDPLY-12247=WebLogic does not support targeting resources to dynamic servers. JRF product related resources \
   will be targeted to the dynamic cluster using the applyJRF function.
 WLSDPLY-12248=The server group(s) {0} will not be targeted to the Admin server or a configured manage server
-WLSDPLY-12249=RCUDBInfo section is specified in the model but the domain type used is not JRF, RCU processing skipped
+WLSDPLY-12249=Domain is not JRF, exiting
```
ddsharpe: but the tool is not exiting, right? Shouldn't this say "skipped"

rakillen: changed to "Domain is not JRF, skipping database configuration"

PR #557, opened 2020-02-26
Add option to update the rcu schema password after the population by …

jshum2479:
Added a flag called -updateRCUSchemaPassword.  When this flag is specified in RCUDBInfo in the domainInfo section, the password will take precedence over the one in the shadow component_schema_info table created by rcu the first time.

### DIFF
--- a/core/src/main/python/update.py
+++ b/core/src/main/python/update.py
@@ -22,6 +22,7 @@ sys.path.append(os.path.dirname(os.path.realpath(sys.argv[0])))
 
 # imports from local packages start here
 from wlsdeploy.aliases.aliases import Aliases
+from wlsdeploy.tool.util.alias_helper import AliasHelper
 from wlsdeploy.aliases.wlst_modes import WlstModes
 from wlsdeploy.exception import exception_helper
 from wlsdeploy.exception.expection_types import ExceptionType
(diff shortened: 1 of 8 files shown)


core/src/main/python/wlsdeploy/util/cla_utils.py
```diff
@@ -70,6 +70,7 @@ class CommandLineArgUtil(object):
     ATTRIBUTES_ONLY_SWITCH     = '-attributes_only'
     FOLDERS_ONLY_SWITCH        = '-folders_only'
     RECURSIVE_SWITCH           = '-recursive'
+    UPDATE_RCU_SCHEMA_PASS_SWITCH = '-updateRCUSchemaPassword'
```
ddsharpe: Why is this new switch necessary?  Why isn't the presence of the new domainInfo attribute sufficient?

core/src/main/python/update.py
```diff
@@ -44,6 +46,20 @@
 from wlsdeploy.util.model import Model
 from wlsdeploy.util.weblogic_helper import WebLogicHelper
 from wlsdeploy.util import model as model_helper
+# from wlsdeploy.aliases.model_constants import RCU_DB_INFO
```
ddsharpe: Why is it necessary to leave all these commented sections of code?  Why not remove them?

core/src/main/python/update.py
```diff
@@ -184,6 +201,81 @@ def __process_encryption_args(optional_arg_map):
         optional_arg_map[CommandLineArgUtil.PASSPHRASE_SWITCH] = String(passphrase)
     return
 
+# def __process_update_rcu_schema_password_args(model, model_context, aliases):
```
ddsharpe: Why is it necessary to leave all these commented sections of code? Why not remove them?

core/src/main/python/wlsdeploy/util/cla_utils.py
jshum2479: This is in WDT which has no idea whether the information is new or old.

core/src/main/python/update.py
jshum2479: I will remove the commented out code

PR #564, opened 2020-03-10
Wdt 343 ds assigned to admin because not targeted

CarolynRountree:
12c - re-arrange extension template code so that configuration of template resources and that the overwrite of domain configuration are successful.


ddsharpe:
Superseded by new PR

PR #574, opened 2020-03-25
Fix operator use case when domain needs to be recreated with updated …

jshum2479:
This fix is to support weblogic kubernetes operator model in image use case when a domain needs to be recreated but the schema password also changed.  We need to setup the shadow table with updated credentials, call get database defaults then reset all the schema passwords before writing the domain


jshum2479:
no need

PR #581, opened 2020-04-02
Remove default for WDT_MODEL_SECRETS_DIRS, skip check if not specified

rakillen:


core/src/main/python/wlsdeploy/util/variables.py
```diff
@@ -357,15 +357,15 @@ def _init_secret_token_map(model_context):
 
     locations = os.environ.get(_secret_dirs_variable, _secret_dirs_default)
     for dir in locations.split(","):
-         if not os.path.isdir(dir):
```
jshum2479, line 360: Do we still wants to have the default set to /weblogic-operator/config-overrides-secrets for public use ?  May be we should remove it ?

PR #587, opened 2020-04-07
Separate validate -print_usage into separate tool

rakillen:
Internal JIRA 388

README updates will be a separate PR.


installer/src/main/bin/validateModel.cmd
```diff
@@ -65,7 +74,6 @@ if "%SHOW_USAGE%" == "false" (
 ECHO.
 ECHO Usage: %SCRIPT_NAME% [-help]
 ECHO              -oracle_home ^<oracle_home^>
-ECHO              [-print_usage ^<context^> [-attributes_only^|-folders_only^|-recursive] ]
 ECHO              [-model_file ^<model_file^>]
```
CarolynRountree, line 77: model file is now required for validate (print usage you did not give model file)

installer/src/main/bin/validateModel.sh
```diff
@@ -31,7 +31,6 @@ usage() {
   echo ""
   echo "Usage: $1 [-help]"
   echo "          -oracle_home <oracle_home>"
-  echo "          [-print_usage <context> [-attributes_only|-folders_only|-recursive] ]"
   echo "          [-model_file <model_file>]"
```
CarolynRountree, line 34: same as above

rakillen: Model can be in the archive file

installer/src/main/bin/validateModel.cmd
rakillen: Model can be in the archive file

PR #592, opened 2020-04-08
Jira wdt 400 version file in zip file

CarolynRountree:
Installs VERSION.txt file in zip file at top of structure. Includes the project.version of the build


installer/pom.xml
```diff
@@ -72,6 +72,29 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>3.1.0</version>
```
ddsharpe, line 77: Plugin version should be derived, not specified.  The parent pom has the resource plugin in the plugin management section.  But, the version may need to be updated.  Currently, the maven-resources-plugin is set to 2.6.

CarolynRountree: It works with both 2.6 and 3.1.0, so left the parent pom at 2.6.

PR #600, opened 2020-04-13
Issue#599 add cdi container mbean

### LINKED ISSUE #599: Ability to set ImplicitBeanDiscoveryEnabled
Hi,

I see that WDT does not discover CdiContainer settings and it does not mention it in source code. We have to set ImplicitBeanDiscoveryEnabled to 0 in CdiContainer folder in Domain.

Is there a way to achieve this or does it require a change by your side?

CarolynRountree:
Fixes #599


core/src/main/python/wlsdeploy/aliases/alias_entries.py
```diff
@@ -87,6 +87,7 @@ class AliasEntries(object):
 
     __topology_top_level_folders = [
         'AdminConsole',
+        'CdiContainer',
```
rakillen, line 90: This will cover createDomain, but will need to be explicitly called out for update and deploy. If it's going to be part of deploy, maybe consider putting it under resources (all three do resources using the same code)

Any other opinions re: resources vs. topology?

PR #601, opened 2020-04-13
Compare Models

jshum2479:
This is a draft PR for the model diff - compareModel tool.   It's a work in progress, there are areas that needs more work.  e.g.  do we allow delete certain resource like connection factory, subedeployment ... from JMSSystemResource -> i.e. generate the right notation at the particular level and what's the best way to accomplish that, right now the tool has no concept of aliases and operated solely on dictionary math.


installer/src/main/bin/compareModel.cmd
```diff
@@ -0,0 +1,73 @@
+@ECHO OFF
+@rem **************************************************************************
+@rem modelHelp.cmd
```
rakillen, line 3: compareModel.cmd

core/src/main/python/model_diff.py
```diff
@@ -0,0 +1,681 @@
...
+        """
+        Update the differences in the final model dictionary with the changes
+        :param ar_changes:   Array of changes in delimited format
+        """
+        # The ar_changes is the keys of changes in the piped format
+        #  'resources|JDBCSystemResource|Generic2|JdbcResource|JDBCConnectionPoolParams|TestConnectionsOnReserve
+        #
+
+        # TODO: may be it can be simplified by walking through levels instead of hard coding?
+
```
rakillen, line 211: I think if you recurse similar to attributes_test.py, you can check for alias_helper.supports_multiple_mbean_instances(), and that will show you which ones have a name level. with apps and libs, I think everything that has a name level can be deleted.

installer/src/main/bin/compareModel.sh
```diff
@@ -0,0 +1,49 @@
...
+# JAVA_HOME             - The location of the JDK to use.  The caller must set
+#                         this variable to a valid Java 7 (or later) JDK.
+#
+
+
+usage() {
+  echo ""
+  echo "Usage: $1 [-help]"
+  echo "          -oracle_home <oracle_home> "
+  echo "          [-compare_model_output_dir <output_dir> write the outputs to the directory specified]"
```
rakillen, line 25: Maybe shorten to -output_dir ? if we do need to reuse that name in the future, cla_utils can deal with it, similar to checks for -domain_home from different tools.

Opinions?

core/src/main/python/model_diff.py
```diff
@@ -0,0 +1,681 @@
...
+        :return:  python dictionary
+        """
+        true = True
+        false = False
+        fh = open(file, 'r')
+        content = fh.read()
+        return eval(content)
+
+
+    def write_dictionary_to_json_file(self, dictionary, writer, indent=''):
```
rakillen, line 482: could PythonToJson in json_translator.py be used for this? might save some duplication.

installer/src/main/bin/compareModel.cmd
```diff
@@ -0,0 +1,73 @@
...
+@rem
+@rem     DESCRIPTION
+@rem       This script compares two models. The models compared must be both yaml or both json
+@rem
+@rem This script uses the following variables:
+@rem
+@rem JAVA_HOME             - The location of the JDK to use.  The caller must set
+@rem                         this variable to a valid Java 7 (or later) JDK.
+@rem
+@rem
```
rakillen, line 19: The variables WLSDEPLOY_HOME and WLSDEPLOY_PROPERTIES are still relevant in jython-only scripts, just copy descriptions from modelHelp.sh and .py

core/src/main/python/model_diff.py
jshum2479: Yes, fixing it

installer/src/main/bin/compareModel.cmd
jshum2479: added

installer/src/main/bin/compareModel.sh
jshum2479: Yes, fixed

installer/src/main/bin/compareModel.cmd
jshum2479: fixed

core/src/main/python/model_diff.py
jshum2479: changed to use alias mechanism to walk the tree.

jshum2479:
I added -variable_file option for resolving tokens

site/compare.md
```diff
@@ -0,0 +1,284 @@
...
+ 3. Deployed two applications `yourear` and `myear` in the appDeployments section
+ 4. Removed application `myear` in the appDeployments section
+ 4. Removed `MDBConnectionFactory` from the `MyJmsModule` JMS Module in the resource section
+ 5. Removed `WebAppContainer` in the resource section
+ 6. Removed the attribute `ResetDeliveryCountOnForward` from `MyUniformDistributedQueue`
+ 7. Changed deployment targets for application `myear`
+ 8. Removed the attribute 'ModuleType' for application `myear` 
+ 
+ **Note**:  The `!` is a notation for the deletion of a non-attribute key element from the model. Any missing attribute will 
+ be omitted from the resulting model, but shown as messages in the output.
```
rosemarymarano, line 274: Any missing attribute will be omitted from the resulting model, -> but shown as a message in the output. 
(all singular OR all plural)
Missing attributes will be omitted from the resulting model, but shown as messages in the output.

PR #623, opened 2020-05-05
Jira wdt 79 catch number format exception

CarolynRountree:


core/src/main/java/oracle/weblogic/deploy/aliases/TypeUtils.java
```diff
@@ -254,6 +255,70 @@ public static Character convertToCharacter(String strValue) {
         }
         return result;
     }
+    private static Integer convertToInteger(String strValue) throws AliasException {
```
ddsharpe, line 258: Too much duplicate code.  Why can't you wrap the IF block with a try and catch the exception once?  Every one of these try-catch blocks is identical.

PR #625, opened 2020-05-06
Documentation for model help -model_sample argument

rakillen:
Documentation tied to #624
Internal JIRA WDT-320


site/model_help.md
```diff
@@ -101,3 +101,40 @@ Path: resources:/JDBCSystemResource
       JDBCXAParams
     SubDeployment (multiple)
 ```
+
+### Model Sample Output
+The `-model_sample` argument can be used to output a model sample for the specified model path. Depending on the output options specified, this argument will create a sample with the available attributes and sub-folders for the specified path.
```
rosemarymarano, line 106: The `-model_sample` argument can be used to output -> You can use the `-model_sample` argument to output

PR #644, opened 2020-06-03
Issue#639 add jpa m bean

CarolynRountree:


core/src/main/python/wlsdeploy/tool/discover/topology_discoverer.py
```diff
@@ -425,6 +427,25 @@ def _get_jmx(self):
...
+        location = LocationContext(self._base_location)
+        location.append_location(model_top_folder_name)
+        name = self._find_singleton_name_in_folder(location)
+        if name is not None:
+            _logger.info('WLSDPLY-06644', class_name=_class_name, method_name=_method_name)
+            location.add_name_token(self._alias_helper.get_name_token(location), name)
+            self._populate_model_parameters(result, location)
+        _logger.exiting(class_name=_class_name, method_name=_method_name)
+        return model_top_folder_name, result
+
```
rakillen, line 448: This is the fifth almost-identical method for discovering a single-unpredictable MBean in this class (sixth if including sub-folders). The only variation is the folder name and error code. Should we create a more general method (preferably in the parent class Discoverer), passing the folder name and using a more general error message, such as "Adding {0}"? Not suggesting a refactor of all discoverer classes, just this one for now.

core/src/main/python/wlsdeploy/aliases/model_constants.py
```diff
@@ -150,6 +150,7 @@
 JMX_NOTIFICATION = 'JMXNotification'
 JNDI_NAME = 'JNDIName'
 JNDI_PROPERTY = 'JNDIProperty'
+JPA = 'JPA'
```
rakillen, line 153: General question - should JPA be included as part of updateDomain? JTA and JMX are excluded, not sure about this one. Opinions?

core/src/main/python/wlsdeploy/tool/discover/topology_discoverer.py
CarolynRountree: I originally wanted to do something a little more complex in my refactor, but will go ahead and start your suggestion. In future, it can be refactored differently.

PR #647, opened 2020-06-10
Jira#wdt 418 encrypt multiple models

CarolynRountree:
Changed to Take a single or multiple models into the Encrypt tool. The passwords in each model are encrypted and each model written out to the original location. The single variable file is shared by the set of models, and the password variables encrypted only once.

Encryption tool did already have checks not to encrypt and encrypted password. I changed the variables code to do a check identical to model code.


ddsharpe:
Please improve description of this change.  Its not clear what you are trying to accomplish with this PR.

installer/src/main/bin/encryptModel.cmd
```diff
@@ -79,11 +79,14 @@ ECHO         oracle_home     - the existing Oracle Home directory for the domain
 ECHO                           This is required unless the ORACLE_HOME environment
 ECHO                           variable is set.
 ECHO.
-ECHO         model_file      - the location of the model file to use.
+ECHO         model_file      - the location and name of the model file to use, or
+ECHO                           the locations and names of each model in a set of
+ECHO                           models. All models will be written back to the same
+ECHO                           location and name.
```
rakillen, line 85: Maybe clarify that comma-separated list for multiples? other tools say "the location of the model file to use.  This can also be specified as a  comma-separated list of model locations, ..."

installer/src/main/bin/encryptModel.sh
```diff
@@ -45,11 +45,14 @@ usage() {
   echo "                          This is required unless the ORACLE_HOME environment"
   echo "                          variable is set."
   echo ""
-  echo "        model_file      - the location of the model file to use."
+  echo "        model_file      - the location and name of the model file or the"
+  echo "                          locations and names of each model in a set of"
+  echo "                          models. All models will be written back to the same"
+  echo "                          location and name."
```
rakillen, line 51: See comment on .sh

Issue #525, opened 2020-01-20
Copy application specific files to domain


larspusa:
Think wlsdeploy archive ZIP might be a good place for application specific files, like MQ bindings files, application property files, xml files etc used by application. Those files are not subject for domainLibraries or classpathLibraries. Files could be copied to domain_home, for example:
wlsdeploy/domainHomeFiles/mqbindings-a/.bindings
wlsdeploy/domainHomeFiles/mqbindings-b/.bindings
wlsdeploy/domainHomeFiles/properties/documentA.properties

To have those files in ZIP archive means those could easily be provided by application provider in DevOps (Jenkins) pipeline and not by infrastructure provider (providing docker wls binary image).

ddsharpe:
@larspusa Is it sufficient to leave these files in a subdirectory of the domain, such as wlsdeploy/additionalFiles?  If so, then we can add an extract to the create domain and update that looks for that same folder in the archive, and extracts it to the domain as wlsdeploy/additionalFiles.

larspusa:
@ddsharpe Would prefer to have it something like i specified, because if we will be starting moving applications one-by-one to docker environments we will not need to investigate if those paths are hardcoded in application or not. I can't say it is a hard requirement from us at the moment though as we still are just doing some research in this technology.
If you at the moment don't think it's a good idea to implement this we can maybe postpone this.

Another question is about .bindings files for ForeignServer. Maybe those should be included in wlsdeploy archive ZIP by default? What do you think?

PR #650, opened 2020-06-15
fix the collection and extraction of foreign connection binding files

CarolynRountree:
Issue #525 requested that WDT collect the MQ Bindings for foreign jms connections. The tool was already collecting, but the resulting MBean attribute in createDomain was not valid. This fix corrects the problem.

The issue requested that the files be collected under domainHome in the archive file, along with other application files. Please review to see if there is additional work to be done.

The MQBindings are collected in wlsdeploy/jms/foreignServer/name of foreign server/

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/deploy/deployer_utils.py
+++ b/core/src/main/python/wlsdeploy/tool/deploy/deployer_utils.py
@@ -5,12 +5,14 @@ Licensed under the Universal Permissive License v 1.0 as shown at https://oss.or
 from sets import Set
 
 from java.io import IOException
+from java.net import URI
 from java.security import NoSuchAlgorithmException
 
 from oracle.weblogic.deploy.util import FileUtils
 from oracle.weblogic.deploy.util import PyWLSTException
 from oracle.weblogic.deploy.util import WLSDeployArchive
 
+from wlsdeploy.aliases.model_constants import FILE_URI
 from wlsdeploy.aliases.wlst_modes import WlstModes
 from wlsdeploy.exception import exception_helper
 from wlsdeploy.exception.expection_types import ExceptionType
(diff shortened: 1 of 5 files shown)


core/src/main/python/wlsdeploy/tool/deploy/deployer_utils.py
```diff
@@ -236,6 +238,39 @@ def ensure_no_uncommitted_changes_or_edit_sessions():
...
+    If the MBean attribute is a URI. To extract it from the archive, need to make
+    it into a relative path.
+    :param model_context: current context containing run information
+    :param path_value: the full value of the path attribute
+    :return: The relative value of the path attribute
+    """
+    uri = URI(path_value)
+    if uri.getScheme() == 'file':
+        value = model_context.tokenize_path(uri.getPath())
+        if value.startswith('@@DOMAIN_HOME@@'):
```
rakillen, line 267: model_context has a constant for this, __DOMAIN_HOME_TOKEN, but it would need to become DOMAIN_HOME_TOKEN (without __).  We could make a method in model_context for this, but it seems very specific to this case.

CarolynRountree: I agree. I was totally annoyed that the constant was private and in the model context instance. Seems it used to be somewhere else but could not find it. I suppose I could take this out of deployer_utils and put in model_context, but I did only leave because of its targeted usage. Votes from anyone else?

ddsharpe: I vote to make it public or a getter if the field needs to be protected for some reason.

CarolynRountree: fixed. made public tokens in model context

PR #655, opened 2020-06-26
Run validator in standalone mode so archive is not required

CarolynRountree:
Changed the compareModel to run the Validator in standalone mode so that the archive file is not required. Since this is not a tool like create or discover, but a utility, there is no reason to run in a stricter mode.  No longer has the SummaryHandler totals which I think is reasonable.

1. Run validator in stand-alone mode
2. Fixed some obvious errors and pe8 warnings

### DIFF
--- a/core/src/main/python/compare_model.py
+++ b/core/src/main/python/compare_model.py
@@ -45,13 +45,13 @@ from wlsdeploy.tool.validate.validator import Validator
 from oracle.weblogic.deploy.validate import ValidateException
 from wlsdeploy.exception.expection_types import ExceptionType
 
-VALIDATION_FAIL=2
-PATH_TOKEN='|'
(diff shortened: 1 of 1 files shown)


core/src/main/python/compare_model.py
```diff
@@ -436,12 +435,12 @@ def compare(self):
 
             variables.substitute(model_dictionary, variable_map, self.model_context)
 
-            return_code = validator.validate_in_tool_mode(model_dictionary,
-                                                      variables_file_name=None,
-                                                      archive_file_name=None)
+            return_code = validator.validate_in_standalone_mode(model_dictionary,
```
rakillen, line 438: This could uses a comment about why, so nobody's tempted to change it back

PR #683, opened 2020-07-22
Wdt 433 compare model validation

CarolynRountree:
1. validation called using model context without a model file. Added code to model_context to copy and add / override arguments. Created a model context copy for each model with the model file in the copied context.
2. compare model to run in validator in stand-alone-mode not completed. This was done becaause tool mode requires an archive file and other more strict policies. The second model file validate was not using stand alone tool mode. and stand alone tool mode not returning a return code from the validate. Fixed both
3. Fixed compare model test file to have something other than a requirement for archive file (not required anymore)

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/validate/validator.py
+++ b/core/src/main/python/wlsdeploy/tool/validate/validator.py
@@ -129,6 +129,7 @@ class Validator(object):
         self._logger.entering(archive_file_name, class_name=_class_name, method_name=_method_name)
         self.__validate_model_file(cloned_model_dict, variable_map, archive_file_name)
 
+        status = Validator.ValidationStatus.VALID
         summary_handler = SummaryHandler.findInstance()
         if summary_handler is not None:
             summary_level = summary_handler.getMaximumMessageLevel()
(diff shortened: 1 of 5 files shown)


core/src/main/python/wlsdeploy/tool/validate/validator.py
```diff
@@ -128,7 +129,20 @@ def validate_in_standalone_mode(self, model_dict, variable_map, archive_file_nam
...
-        self._logger.exiting(class_name=_class_name, method_name=_method_name)
+        summary_handler = SummaryHandler.findInstance()
+        if summary_handler is not None:
+            summary_level = summary_handler.getMaximumMessageLevel()
+            if summary_level == Level.SEVERE:
+                status = Validator.ValidationStatus.INVALID
+            elif summary_level == Level.WARNING:
+                status = Validator.ValidationStatus.WARNINGS_INVALID
+
+        status = Validator.ValidationStatus.VALID
```
rakillen, line 140: Should status be set to VALID **before** checking handler?
Here it will always make the return code PROCEED.

PR #692, opened 2020-07-30
Fix for top level removal between models

CarolynRountree:
Did a lot of testing and found only the removal of a top level causes the problem. Problem due to an undefined variable.

Fixes Jira WDT-436

### DIFF
--- a/core/src/test/python/compare_model_test.py
+++ b/core/src/test/python/compare_model_test.py
@@ -190,6 +190,60 @@ class CompareModelTestCase(unittest.TestCase):
 
         self.assertNotEqual(return_code, 0)
 
+    def testDeleteModelAppDeploymentsRemoved(self):
+        _method_name = 'testCompareModelFull'
+
+        _variables_file = self._resources_dir + '/compare_model_model1.10.properties'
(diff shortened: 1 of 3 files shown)


core/src/test/python/compare_model_test.py
```diff
@@ -190,6 +190,60 @@ def testCompareModelInvalidFile(self):
 
         self.assertNotEqual(return_code, 0)
 
+    def testDeleteModelAppDeploymentsRemoved(self):
```
ddsharpe, line 193: Is this like a double negative?  Delete ... Removed.  Please add a comment, or change the name of the method to clarify the intent.

core/src/test/python/compare_model_test.py
```diff
@@ -190,7 +190,7 @@ def testCompareModelInvalidFile(self):
 
         self.assertNotEqual(return_code, 0)
 
-    def testDeleteModelAppDeploymentsRemoved(self):
+    def testDeleteModelAppDeployments(self):
         _method_name = 'testCompareModelFull'
```
rakillen, line 194: method name doesn't match

PR #712, opened 2020-08-21
Verify that sections are dictionaries, and not single values

rakillen:
When a group is expected in the model, and a string is found, a stack trace occurs. This model has an indention problem, and and `Target` appears to be a JDBC name. The tools fail with a stack trace, expecting it to have a nested group instead of string.

    JDBCSystemResource:
        'Bubba-DS2':
        Target: mycluster
        JdbcResource:
stack trace snippet:

 File "/u01/wdt/weblogic-deploy/lib/python/wlsdeploy/tool/validate/validator.py", line 384, in _Validator__validate_model_section
 File "/u01/wdt/weblogic-deploy/lib/python/wlsdeploy/tool/validate/validator.py", line 461, in _Validator__validate_section_folder
 File "/u01/wdt/weblogic-deploy/lib/python/wlsdeploy/tool/validate/validator.py", line 532, in _Validator__process_model_node
AttributeError: 'string' object has no attribute 'iteritems'

### DIFF
--- a/core/src/main/resources/oracle/weblogic/deploy/messages/wlsdeploy_rb.properties
+++ b/core/src/main/resources/oracle/weblogic/deploy/messages/wlsdeploy_rb.properties
@@ -366,7 +366,7 @@ WLSDPLY-01803=Adding roles: {0}
 WLSDPLY-01804=Failed to convert role {0} with expression {1} because {2}
 WLSDPLY-01805=Unexpected {0} during role mapper processing: {1}
 
-# wlsdeploy/util/mbean_utils.py
+# wlsdeploy/tool/util/credential_map_helper.py
 WLSDPLY-01790=Creating default credential mapper initialization file {0}
 WLSDPLY-01791=Attribute "{0}" is required for {1} credential mapping {2}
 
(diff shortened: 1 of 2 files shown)


core/src/main/resources/oracle/weblogic/deploy/messages/wlsdeploy_rb.properties
```diff
@@ -366,7 +366,7 @@ WLSDPLY-01803=Adding roles: {0}
 WLSDPLY-01804=Failed to convert role {0} with expression {1} because {2}
 WLSDPLY-01805=Unexpected {0} during role mapper processing: {1}
 
-# wlsdeploy/util/mbean_utils.py
+# wlsdeploy/tool/util/credential_map_helper.py
```
CarolynRountree, line 369: why commented out?

rakillen: This is a correction from my last PR - I copy/pasted the wrong heading to a section with new messages for credential mapper

PR #713, opened 2020-08-21
Changes for Verrazzano

ddsharpe:
* added new target type, wko
* added execute permissions to generated shell script from vz target
* corrected weblogic credential secret name for models targeted to WebLogic Operator
* changed generated Verrazzano binding defaults for DNS name to "*"
* changed generated Verrazzano binding namepace from default to {domainUid}-ns
* corrected log folder name from jcslcm-logs to wdt-logs

### DIFF
--- a/core/src/test/python/target_configuration_helper_test.py
+++ b/core/src/test/python/target_configuration_helper_test.py
@@ -1,0 +1,56 @@
+"""
+Copyright (c) 2020, Oracle Corporation and/or its affiliates.  All rights reserved.
+The Universal Permissive License (UPL), Version 1.0
+"""
+
+import unittest
+
+import wlsdeploy.util.target_configuration_helper as HELPER
+import wlsdeploy.util.target_configuration as CONFIG
+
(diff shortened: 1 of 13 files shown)


core/src/test/python/target_configuration_helper_test.py
```diff
@@ -0,0 +1,56 @@
...
+    target_with_cred_name = None
+    target_without_cred_name = None
+
+    def setUp(self):
+        config = dict()
+        config[CONFIG.CREDENTIALS_METHOD] = 'secrets'
+        config[CONFIG.WLS_CREDENTIALS_NAME] = '__weblogic-credentials__'
+        self.target_with_cred_name = TargetConfiguration(config)
+
+        config = dict()
```
jshum2479, line 29: duplicate with lines 24-27?

ddsharpe: two different configs.  changed the temporary variable names to avoid any sharing between classes during the test.

PR #714, opened 2020-08-21
Document user password credential mapping

rakillen:


site/use_cases.md
```diff
@@ -409,7 +410,43 @@
  - WebLogic global roles are only updated for the WebLogic security XACML role mapping provider (i.e. `XACMLRoleMapper`)
  - The user and group processing is not complete, currently, users cannot be assigned to groups. Users created using the `Security` section are automatically added to the `Administrators` group and are not added to the groups specified. As soon as a patch to correct the user and group processing is available, we will post it here.
 
-
+ ### Modeling WebLogic User Password Credential Mapping
+ 
+ The Create Domain tool can be used to create user password credential mappings. Information in the model will be used to create a credential mapping file that will be imported the first time the Administration Server is started. This example shows how mappings are represented in the model:
```
rosemarymarano, line 415: Create Domain tool -> Create Domain Tool

site/use_cases.md
```diff
@@ -409,7 +410,43 @@
...
+                Method: GET
+                User: 'user1,user2'
+                RemoteUser: remoteUser2
+                RemotePassword: '@@PROP:remote2.pwd@@'
+```
+ In this example, the mapping `map1` creates a cross-domain credential mapping that provides access from this domain to the remote domain `otherDomain` as the user `otherUser` with the configured password.
+ 
+ The mapping `map2` creates a remote resource credential mapping that will give the local user `user1` access to a single remote resource on `remote.host` as the user `remoteUser` with the configured password. The mapping `map3` is similar, but provides access to a different remote resource for two local users, `user1` and `user2`.
+ 
+ The names of the mapping sections in the model, such as `map1` and `map2`, are used to group the attributes for each mapping in the model, and are not part of the resulting credential mappings. These names should be unique for each mapping of a particular type.
```
rosemarymarano, line 448: model, -> model (no comma needed here)

PR #723, opened 2020-08-28
Tokenize user credentials in model, associate with passwords for secrets

rakillen:
credential_injector.py is a new subclass of variable_injector that is used by discoverDomain and prepareModel to collect credential fields as the model is traversed. It is extended to use secret names as keys in its internal cache, and to inject secret tokens in the model when the target type uses secrets.

Processing to tokenize JDBC user properties and mail session credentials was incorporated. 

discoverDomain already used a variable injector for this purpose. A lot of changes were made to just rename variable_injector to credential_injector throughout.

prepareModel was modified (a lot) to use the credential injector, and remove its duplicate code in this area. Verified with multiple model files.

The k8s secrets creation script uses the credential injector's cache keys to combine user and password keys into a single secret.

Internal Jira WDT-465


ddsharpe (approved):
ran several tests, LGTM

CarolynRountree (approved):

CarolynRountree (approved):

PR #725, opened 2020-09-02
Generate a shortened name when secret name exceeds 63 chars

CarolynRountree:
If name exceeds secret name length (currently 63 chars), truncate the name and overlay the last 5 chars of the truncated name with a 5 char number. 

Log a warning about the truncation.

### DIFF
--- a/core/src/main/python/wlsdeploy/util/target_configuration_helper.py
+++ b/core/src/main/python/wlsdeploy/util/target_configuration_helper.py
@@ -274,9 +274,9 @@ def _create_secret_name(variable_name, secret_length=SECRET_NAME_LENGTH):
     # if empty, just return "x".
     secret = '-'.join(secret_keys).strip('-')
     if len(secret) > secret_length:
-        __logger.warning('WLSDPLY-07000', secret_length, secret, class_name=__class_name, method_name=_method_name)
         secret = secret[:secret_length] + '%05d' % __long_key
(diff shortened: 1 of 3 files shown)


core/src/main/python/wlsdeploy/util/target_configuration_helper.py
```diff
@@ -269,6 +273,11 @@ def _create_secret_name(variable_name):
     # rejoin with hyphens, remove leading and trailing hyphens from final name.
     # if empty, just return "x".
     secret = '-'.join(secret_keys).strip('-')
+    if len(secret) > secret_length:
+        secret = secret[:secret_length] + '%05d' % __long_key
```
jshum2479, line 277: The max is 63 characters and adding 5 digits will be longer than 63

core/src/main/python/wlsdeploy/util/target_configuration_helper.py
```diff
@@ -269,6 +273,11 @@ def _create_secret_name(variable_name):
...
+        __long_key += 1
```
rakillen, line 278: I think this will result in the secret names in the script having different numbers than the ones in the model, as this method is called again for each variable/secret name. If this fix is applied with PR #723, this is resolved.
This method is called again when the template hashes are generated for target vz, even with #723, so it may need to be revised to use the credential cache.

PR #727, opened 2020-09-09
Allow AppDir in model domainInfo section

rakillen:
Was failing validation, not in modelHelp.

Set AppDir option regardless of extension templates


jshum2479 (approved):
LGTM

CarolynRountree (approved):

PR #731, opened 2020-09-16
Allow user to undo any current edit session

jshum2479:
Added parameter -undo_current_edit for online update and deploy.  If the parameter is present, any outstanding unactivated changes will be discarded before proceeding.


core/src/main/python/update.py
```diff
@@ -125,9 +126,11 @@ def __update_online(model, model_context, aliases):
 
     try:
         __wlst_helper.connect(admin_user, admin_pwd, admin_url)
-        deployer_utils.ensure_no_uncommitted_changes_or_edit_sessions()
+        deployer_utils.ensure_no_uncommitted_changes_or_edit_sessions(model_context.is_undo_current_edit())
         __wlst_helper.edit()
         __wlst_helper.start_edit()
+        if model_context.is_undo_current_edit():
+            deployer_utils.undo_current_edit()
```
rakillen, line 133: Does undo_current_edit need to happen before the new edit is started?

core/src/main/python/deploy.py
```diff
@@ -121,7 +122,7 @@ def __deploy_online(model, model_context, aliases):
 
     try:
         __wlst_helper.connect(admin_user, admin_pwd, admin_url)
-        deployer_utils.ensure_no_uncommitted_changes_or_edit_sessions()
+        deployer_utils.ensure_no_uncommitted_changes_or_edit_sessions(model_context.is_undo_current_edit())
```
rakillen, line 125: Deploy should do check and do deployer_utils.undo_current_edit same as update.

core/src/main/python/update.py
jshum2479: Yes, it needs to have an edit session first

core/src/main/python/deploy.py
jshum2479: added

rakillen (approved):

CarolynRountree (approved):

ddsharpe (approved):

PR #734, opened 2020-09-22
Fix compareModel stack dump when variables file not specified

rakillen:
Pass empty variables map to validate to prevent crash (lines 458, 477)

Optimized imports (no changes).

A lot of whitespace changes to remove PEP-8 warnings.

Internal JIRA WDT-467

### DIFF
--- a/core/src/main/python/compare_model.py
+++ b/core/src/main/python/compare_model.py
@@ -14,39 +14,41 @@
 #
 
 import sets
-import sys, os, traceback
+import traceback
 
-from java.lang import System
 import java.io.File as JFile
 import java.io.FileOutputStream as JFileOutputStream
 import java.io.IOException as JIOException
 import java.io.PrintWriter as JPrintWriter
-import oracle.weblogic.deploy.util.TranslateException as TranslateException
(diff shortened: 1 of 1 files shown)


ddsharpe (approved):

core/src/main/python/compare_model.py
```diff
@@ -14,39 +14,41 @@
...
-import sys, os, traceback
+import traceback
 
-from java.lang import System
 import java.io.File as JFile
 import java.io.FileOutputStream as JFileOutputStream
 import java.io.IOException as JIOException
 import java.io.PrintWriter as JPrintWriter
-import oracle.weblogic.deploy.util.TranslateException as TranslateException
+import os
```
CarolynRountree, line 23: should be separate from java exceptions

CarolynRountree (changes requested):

core/src/main/python/compare_model.py
rakillen: Moved back up with sets, traceback

CarolynRountree (approved):

PR #736, opened 2020-09-24
Enable ODL configuration for dynamic servers

rakillen:
Use logging configuration template to configure ODL when server has not been created.


core/src/main/python/wlsdeploy/tool/deploy/odl_deployer.py
```diff
@@ -100,10 +102,24 @@ def _update_server(self, name, dictionary, config_location):
...
+                if not server_dir.exists() and not server_dir.mkdirs():
+                    self.logger.severe('WLSDPLY-19707', server_dir, class_name=self.__class_name,
+                                       method_name=_method_name)
+                    return
+                FileUtils.validateWritableDirectory(server_dir.getPath())
+
+            document = LoggingConfigurationDocument(FileInputStream(source_file))
 
             # configure AddJvmNumber
             add_jvm_number = dictionary_utils.get_element(dictionary, _ADD_JVM_NUMBER)
```
jshum2479, line 125: There is try except section but when there is an exception. It only logs but not throwing an exception, I think it should throw an exception.

jshum2479 (approved):

ddsharpe (approved):

PR #745, opened 2020-10-07
Don't capitalize true/false values when parsing model

rakillen:
Example: WebAppFilesCaseInsensitive can have values of [false, os, true]

Internal Jira WDT-463

### DIFF
--- a/core/src/main/java/oracle/weblogic/deploy/yaml/AbstractYamlTranslator.java
+++ b/core/src/main/java/oracle/weblogic/deploy/yaml/AbstractYamlTranslator.java
@@ -305,10 +305,8 @@ public abstract class AbstractYamlTranslator extends YamlBaseListener {
         String booleanValue = "False";
         if (!StringUtils.isEmpty(text)) {
             text = StringUtils.stripQuotes(text.trim());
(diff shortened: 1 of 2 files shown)


core/src/main/java/oracle/weblogic/deploy/yaml/AbstractYamlTranslator.java
```diff
@@ -305,10 +305,11 @@ private PyObject getBooleanValue(String name, String text) {
         String booleanValue = "False";
         if (!StringUtils.isEmpty(text)) {
             text = StringUtils.stripQuotes(text.trim());
+            // don't adjust case, these may not be intended as boolean
             if ("true".equalsIgnoreCase(text)) {
-                booleanValue = "True";
+                booleanValue = text;
```
ddsharpe, line 310: Maybe combine the two branches of this IF since they are the same?

rakillen: done

ddsharpe (approved):

CarolynRountree (approved):

PR #787, opened 2020-11-18
Morecomparetest

jshum2479:
Remove global use in compare_model.py and add more compare models test


rakillen (approved):

rakillen:
I added a comment to WDT-513 to re-enable _testCompareModelFull3 when it is fixed.

CarolynRountree (approved):

Issue #784, opened 2020-11-17
Discovery domain online - WLSDPLY-00005: get(null) in online mode failed


textanalyticsman:
Hello,

I am trying to discover a domain online and I am getting errors detailed below.

I am using this command.

`/home/soa/weblogic-deploy/bin/discoverDomain.sh -oracle_home /u01/binaries/soa/fmw/12.1.3.0.0 -domain_home /u01/soa/domain/soa_domain -archive_file /home/soa/SOADomain.zip -model_file /home/soa/SOADomainModel.yaml -domain_type SOA -java_home /u01/binaries/soa/jdk/current -variable_file /home/soa/SOADomainModel.properties -admin_user weblogic -admin_url t3://localhost:7011 -wlst_path /u01/binaries/soa/fmw/12.1.3.0.0/soa`

The domain I am trying to introspect is a SOA domain 12.1.3.

####<Nov 17, 2020 7:15:58 PM> <FINER> <WlstHelper> <get> <> <THROW
oracle.weblogic.deploy.discover.DiscoverException: WLSDPLY-00005: get(null) in online mode failed: Error occurred while performing get : Error getting attribute null : Exception occured trying to get attribute of a ModelMBean 
Use dumpStack() to view the full stacktrace :w
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
	at org.python.core.PyReflectedConstructor.__call__(Unknown Source)
	at org.python.core.PyJavaInstance.__init__(Unknown Source)
	at org.python.core.PyJavaClass.__call__(PyJavaClass.java:891)
	at org.python.core.PyObject.__call__(Unknown Source)
	at wlsdeploy.exception.exception_helper$py.create_discover_exception$6(/home/elhsoa/weblogic-deploy/lib/python/wlsdeploy/exception/exception_helper.py:150)
	at wlsdeploy.exception.exception_helper$py.call_function(/home/elhsoa/weblogic-deploy/lib/python/wlsdeploy/exception/exception_helper.py)
	at org.python.core.PyTableCode.call(Unknown Source)
[...]

PR #789, opened 2020-11-19
Wdt 516 dynamic cluster size

CarolynRountree:
Fixes Issue #784


core/src/main/python/wlsdeploy/aliases/model_constants.py
```diff
@@ -175,6 +175,7 @@
 MAIL_SESSION = 'MailSession'
 MAIL_SESSION_OVERRIDE = 'MailSessionOverride'
 MAIL_SESSION_PROPERTIES = 'Properties'
+MAX_DYNAMIC_CLUSTER_SIZE = 'MaximumDynamicServerCount'
```
rakillen, line 178: it's minor, but can you make the variable name reflect the value? cluster has too many similar attributes related to dynamic server count.

rakillen (approved):

core/src/main/python/wlsdeploy/tool/discover/topology_discoverer.py
```diff
@@ -768,7 +768,7 @@ def _dynamic_server(self, server, location):
                     _logger.fine('WLSDPLY-06613', server, class_name=_class_name, method_name=_method_name)
                     self._wlst_helper.cd(wlst_path)
                     attr_name = self._aliases.get_wlst_attribute_name(cluster_location,
-                                                                      model_constants.DYNAMIC_CLUSTER_SIZE)
+                                                                          model_constants.MAX_DYNAMIC_CLUSTER_SIZE)
```
jshum2479, line 771: It should work but the MaximumDynamicServerCount is deprecated  

@Deprecated
int getMaximumDynamicServerCount()
Deprecated. 12.2.1.0.0 Use getDynamicClusterSize()
The maximum number of server instances allowed to be created. This value should be set to the number of servers expected to be in the cluster. The specified number of servers are dynamically added to the configuration at runtime and associated ServerLifeCycleRuntimeMBeans are created.
Returns:
maximum number of dynamic servers to be created
Changes to this property are dynamic and thus take effect immediately without restarting the server or partition.
Maximum Value:
800
Minimum Value:
0

CarolynRountree: changed

ddsharpe (approved):

rakillen (approved):

PR #801, opened 2020-12-04
Compare model: compare simple properties correctly

rakillen:
Fix output comments.
Enable unit test with minor revisions.

Internal Jira WDT-512

### DIFF
--- a/core/src/main/python/compare_model.py
+++ b/core/src/main/python/compare_model.py
@@ -34,6 +34,7 @@ from oracle.weblogic.deploy.validate import ValidateException
 import oracle.weblogic.deploy.util.TranslateException as TranslateException
 from wlsdeploy.aliases import alias_utils
 from wlsdeploy.aliases.alias_constants import ALIAS_LIST_TYPES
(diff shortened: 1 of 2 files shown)


core/src/main/python/compare_model.py
```diff
@@ -430,6 +443,8 @@ def _get_change_info(self, current_value, previous_value, location, attribute_na
                 current_text = ','.join(current_list)
                 previous_text = ','.join(previous_list)
                 comment = attribute_name + ": '" + previous_text + "' -> '" + current_text + "'"
+            elif attribute_type == PROPERTIES:
+                comment = property_key + ": '" + str(previous_value) + "'"
             elif not isinstance(previous_value, dict):
                 comment = attribute_name + ": '" + str(previous_value) + "'"
```
jshum2479, line 449: I think we should mask the password value

rakillen: The password from the new model needs to be exposed in this output, correct?
Are you recommending that we mask the comment with the password from the previous model only?

jshum2479 (approved):

ddsharpe (approved):

PR #810, opened 2020-12-16
Set domain source type default for extractResource and target resource files

rakillen:
Updated modelHelp to show enum values where available.
Changes made in the `JIRA-WDT-522-vz-oam-descriptors` branch as well.

Internal Jira WDT-530

### DIFF
--- a/core/src/main/targetconfigs/vz/model.yaml
+++ b/core/src/main/targetconfigs/vz/model.yaml
@@ -13,6 +13,10 @@ spec:
         # The WebLogic Domain Home
         domainHome: {{{domainHome}}}
 
+        # The domain home source type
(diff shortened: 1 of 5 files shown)


CarolynRountree (approved):

core/src/main/targetconfigs/vz/model.yaml
```diff
@@ -13,6 +13,10 @@ spec:
         # The WebLogic Domain Home
         domainHome: {{{domainHome}}}
 
+        # The domain home source type
+        # Set to PersistentVolume for domain-in-pv, Image for domain-in-image, or FromModel for model-in-image
+        domainHomeSourceType: {{{domainHomeSourceType}}}
```
ddsharpe, line 18: This requires a change to WebLogic Image Tool.  It will not break, but it will not be populated.

ddsharpe (approved):

PR #816, opened 2020-12-24
WIP: Avoid undeploying apps and libraries when targets are deleted

rakillen:
Delete app/library targets during edit session for online.
Remove target deletion functionality from online app/library deployment.
Refactored restart checking into a shared method online_check_save_activate().
Tested with update and deploy tools.

Internal Jira WDT-534


jshum2479 (approved):
LGTM

ddsharpe (approved):

PR #820, opened 2021-01-13
Avoid re-deploying clusters and servers a second time

### LINKED ISSUE #819: Different behaviour 1.9.8 vs 1.9.7 in create mode
After upgrading to 1.9.8 our standard create domain scripts failed, while configuration was left identical. We returned to use 1.9.7 successfully, analysing the issue later:

Until **1.9.7**. is was possible to define the SSL port and disable it all together.
<wdt config snippet>
```
      SSL:
        ListenPort: 7600
        Enabled: false
```
Running **createDomain** would define the portnumber but disable the SSL like we designed.
(it is out of scope why we designed it this way)

From **1.9.8** running **createDomain** with this same configuration raises:
```
WLSDPLY-**12409**: createDomain failed to create the domain: Failed to set attribute ListenPort in path /Server/AS_FRS12214/SSL/NO_NAME_0 to value 7,600: 64016: Attempt to set disabled property.
64016: The property "SSL Listen Port" is disabled on type SSL. 
```
and the same behaviour in **updateDomain**, though different error code:
```
WLSDPLY-**09015**: updateDomain deployment failed: Failed to set attribute ListenPort in path /Server/AS_FRS12214/SSL/AS_FRS12214 to value 7,600: 64016: Attempt to set disabled property.
64016: The property "SSL Listen Port" is disabled on type SSL.
```
_(with hinesight it seemed that WLSDPLY-09015 was also raised in **1.9.7 updateDomain**, but we never used thatone before)_

Would like to understand/know if this was our design flaw, or a change in WDT

rakillen:
After setting server groups, just remove the deleted items.
WLST does not allow SSL/ListenPort to be set a second time.

Also fixed PyDictionary type match for new Jython version.

Fixes #819

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/create/domain_creator.py
+++ b/core/src/main/python/wlsdeploy/tool/create/domain_creator.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2021, Oracle and/or its affiliates.
+Copyright (c) 2017, 2021, Oracle and/or its affiliates.
 Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 """
 import os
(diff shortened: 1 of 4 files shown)


CarolynRountree (approved):

ddsharpe (approved):
LGTM

PR #821, opened 2021-01-14
Compare tool: keep source path, don't delete non-name folders

rakillen:
Keep the source path for applications and libraries in the change model, even if unchanged.
Internal Jira WDT-535

Avoid flagging non-named model folders for deletion in the change model.
Internal Jira WDT-527

ModelDiffer has been moved out of compare_model.py (see model_comparer.py).
Refactored as ModelComparer to use a single pass through the models, and provide additional checks.

Some messages in applications_deployer.py were corrected to indicate Application or Library as needed.
Fix in alias_utils.py for merging properties.

### DIFF
--- a/core/src/main/python/compare_model.py
+++ b/core/src/main/python/compare_model.py
@@ -281,6 +281,7 @@ def main():
             if len(obj.get_compare_msgs()) > 0:
                 try:
                     file_name = _outputdir + '/compare_model_stdout'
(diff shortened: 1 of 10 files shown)


ddsharpe (approved):
LGTM but do not have enough understanding of the use case to test.  Defering testing to Johnny before merging.

jshum2479 (approved):

PR #828, opened 2021-01-29
Allow discover to run without required archive file

### LINKED ISSUE #824: Request feature discoverDomain.sh optional usage of -archive_file option
I'd like to request a feature that could be interesting for more users: optional usage of -archife_file option. The motivation is for many purposes the archive is not needed. For example, to review the domain configuration or modify most of the parameters you just need the model file. However, gathering the archive_file is a process that could take several minutes. I do have domains where the process of creating the archive file takes an hour. That is very inconvenient. 

Thank you for evaluating the proposal.

CarolynRountree:
Fixes #824


rakillen (approved):

core/src/main/python/discover.py
```diff
@@ -94,35 +94,52 @@ def __process_args(args):
...
+            CommandLineArgUtil.MODEL_FILE_SWITCH not in argument_map:
+        ex = exception_helper.create_cla_exception('WLSDPLY-06028')
+        __logger.throwing(ex, class_name=_class_name, method_name=_method_name)
+        raise ex
+
+
+def __process_archive_filename_arg(optional_arg_map):
     """
     Validate the archive file name and load the archive file object.
-    :param required_arg_map: the required arguments map
```
rakillen, line 107: Very minor, but the map passed in includes required and optional arguments after CLA processing, maybe call it `argument_map` like above method.

core/src/main/java/oracle/weblogic/deploy/util/WLSDeployArchive.java
```diff
@@ -201,9 +228,11 @@ public void addModel(File model) throws WLSDeployArchiveIOException {
         final String METHOD = "addModel";
 
         LOGGER.entering(CLASS, METHOD, model);
-        validateExistingFile(model, "model", getArchiveFileName(), METHOD);
-        getZipFile().removeZipEntries(ARCHIVE_MODEL_TARGET_DIR);
-        addItemToZip(ARCHIVE_MODEL_TARGET_DIR, model);
+        if (hasArchive) {
```
ddsharpe, line 231: With a code block this small, it doesn't matter much.  But, code analyzers tend to report that the quick test and return are less "complex" for the reader.  Like
if (hasArchive) {
  return;
}

ddsharpe (approved):
LGTM

core/src/main/java/oracle/weblogic/deploy/util/WLSDeployArchive.java
CarolynRountree: I would agree with that except that all of these methods have enter / exit logging

robertpatrick (approved):

PR #838, opened 2021-02-10
remove WeblogicHelper dependencies from aliases for aliastest

CarolynRountree:
Remove hard dependencies on Weblogic classes from WeblogicHelper in the aliases and alias_entries python files. This will allow the alias test to run without having a dependency on an oracle home.

### DIFF
--- a/core/src/main/python/wlsdeploy/util/string_utils.py
+++ b/core/src/main/python/wlsdeploy/util/string_utils.py
@@ -1,9 +1,10 @@
 """
-Copyright (c) 2017, 2019, Oracle Corporation and/or its affiliates.  All rights reserved.
+Copyright (c) 2017, 2021, Oracle Corporation and/or its affiliates.  All rights reserved.
 Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 
 This module provides string manipulation helper methods that are not found in the WLST version of Jython
 """
+import java.lang.String as JString
 import java.util.Properties as Properties
 import java.io.FileInputStream as FileInputStream
 import java.io.IOException as IOException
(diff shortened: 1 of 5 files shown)


core/src/main/python/wlsdeploy/util/string_utils.py
```diff
@@ -104,3 +107,56 @@ def load_properties(property_file, exception_type=None):
         prop_dict[key] = value
 
     return prop_dict
+
+
+def is_weblogic_version_or_above(wls_version, str_version):
+    """
+    Is the provided version number equal to or greater than the version encapsualted by this version instance
+    :param wls_version: the array representation of the current weblogic version to be compared
+    :param str_version: the string representation of the version to be compared
```
rakillen, line 116: I think both arguments are strings (both call .split()), also hard to describe. maybe:
```
wls_version: the string representation of the WebLogic version to be compared
str_version: the string representation of the WebLogic version to compare against
```
?

core/src/main/python/wlsdeploy/util/string_utils.py
```diff
@@ -104,3 +107,56 @@ def load_properties(property_file, exception_type=None):
         prop_dict[key] = value
 
     return prop_dict
+
+
+def is_weblogic_version_or_above(wls_version, str_version):
+    """
+    Is the provided version number equal to or greater than the version encapsualted by this version instance
```
rakillen, line 114: hard to describe, maybe:
```
... than the version in the second argument.
```
(same for :return:)

core/src/main/python/wlsdeploy/util/string_utils.py
CarolynRountree: Thanks changed.

rakillen (approved):

ddsharpe (approved):

PR #840, opened 2021-02-17
For -target vz and wko, omit cluster section if list is empty

rakillen:
Problems discussed in verrazzano-dev Slack:
- Section `clusters` should be omitted if list is empty.
- A control character was removed from application.yaml .

For template processing, skip template block if false value in hash.
Internal Jira WDT-522


core/src/main/targetconfigs/vz/application.yaml
```diff
@@ -1,6 +1,6 @@
 # Copyright (c) 2020, 2021, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
-​
+
```
rakillen, line 3: this doesn't show in review, but removing `<200b>` control character on this line.

ddsharpe (approved):

CarolynRountree (approved):

PR #852, opened 2021-03-14
Filtering on validate

lennyphan:
To support generating configuration overrides with model filtering , for Model-In-Image (MII) domain home source type, the following changes are needed in WDT:

1. Add the ability to apply filters for The Validate Model Tool (validateModel),
2. Persist model after filtering has been applied.

For MII, validateModel tool is used to generate merged models and filtering is needed to apply configuration overrides.

A couple unit tests created to verify that model filter(s) are invoked (for validateModel tool) and to verify that filter changes are persisted, if configured.

### DIFF
--- a/core/src/main/python/validate.py
+++ b/core/src/main/python/validate.py
@@ -111,12 +111,6 @@ def __perform_model_file_validation(model_file_name, model_context):
         variable_map = model_validator.load_variables(model_context.get_variable_file())
         model_dictionary = cla_helper.merge_model_files(model_file_name, variable_map)
 
-        # apply filters to merged model
-        if variable_map:
-            # if there are any variables than substitute them in the model
-            variables.substitute(model_dictionary, variable_map, model_context)
-        filter_helper.apply_filters(model_dictionary, "validate")
-
         if cla_helper.check_persist_model():
             persist_model_dict = copy.deepcopy(model_dictionary)
             variables.substitute(persist_model_dict, variable_map, model_context)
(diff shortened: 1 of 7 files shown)


jshum2479 (approved):
LGTM

ddsharpe (approved):

CarolynRountree (approved):

CarolynRountree (approved):

core/src/main/python/validate.py
```diff
@@ -110,6 +111,12 @@ def __perform_model_file_validation(model_file_name, model_context):
         variable_map = model_validator.load_variables(model_context.get_variable_file())
         model_dictionary = cla_helper.merge_model_files(model_file_name, variable_map)
 
+        # apply filters to merged model
+        if variable_map:
+            # if there are any variables than substitute them in the model
+            variables.substitute(model_dictionary, variable_map, model_context)
```
rakillen, line 117: the model should still have variables unresolved before going into `validate_in_standalone_mode`, so that any missing variables can be logged there.

rakillen (changes requested):
the model should still have variables unresolved before going into validate_in_standalone_mode, so that any missing variables can be logged there.

core/src/main/python/validate.py
lennyphan: The change now invokes filtering after validating the merged models (i.e calling **validate_in_standalone_mode**).  We will validate a second time (and re-persist) if filters were applied.

rakillen (approved):

PR #865, opened 2021-04-02
Throw an exception when missing security provider schematype jar

CarolynRountree:
This was requested by Johnny Shum due to the problems troubleshooting in wit. Changed from warning to exception, and job fails for any security provider discovered in offline that does not have a schematype jar. Particularly the trust identity asserter.


CarolynRountree:
Something is off with this PR. Closing and opening another

ddsharpe (approved):

jshum2479 (approved):

ddsharpe:
Leaving missing JAR as a WARNING, and closing this PR for now.  Discovery was not meant to "break" when something wasn't found which leads me to believe that the warning is the correct result.

PR #867, opened 2021-04-05
fix empty target for online deploy/undeploy

jshum2479:
Undeploy the app if all targets are removed from the target for online update

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/deploy/applications_deployer.py
+++ b/core/src/main/python/wlsdeploy/tool/deploy/applications_deployer.py
@@ -643,6 +643,11 @@ class ApplicationsDeployer(Deployer):
                                                                                     from_archive=True)
 
                 existing_app_ref = dictionary_utils.get_dictionary_element(existing_app_refs, versioned_name)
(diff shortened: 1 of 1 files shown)


jshum2479:
closing for now

core/src/main/python/wlsdeploy/tool/deploy/applications_deployer.py
```diff
@@ -643,6 +643,11 @@ def __build_app_deploy_strategy(self, location, model_apps, existing_app_refs, s
                                                                                     from_archive=True)
 
                 existing_app_ref = dictionary_utils.get_dictionary_element(existing_app_refs, versioned_name)
+                if len(dictionary_utils.get_element(app_dict, TARGET)) == 0:
```
ddsharpe, line 646: What does get_element return if TARGET is not present?  Does not present get the same result as zero length string?

jshum2479: This PR has been closed and no longer needed

PR #874, opened 2021-04-19
Fix to resolve env in key before merging

jshum2479:
This fix is for JIRA WDT 552 - resolving the key in the model with @@ENV notation before merging the models to avoid having duplicate copies of the same key element in the model resulting wrong domain created.

### DIFF
--- a/core/src/main/python/wlsdeploy/util/cla_helper.py
+++ b/core/src/main/python/wlsdeploy/util/cla_helper.py
@@ -228,7 +228,7 @@ def load_model(program_name, model_context, aliases, filter_type, wlst_mode):
 
     model_file_value = model_context.get_model_file()
     try:
-        model_dictionary = merge_model_files(model_file_value, model_context, variable_map)
+        model_dictionary = merge_model_files(model_file_value, variable_map)
     except TranslateException, te:
         __logger.severe('WLSDPLY-09014', program_name, model_file_value, te.getLocalizedMessage(), error=te,
                         class_name=_class_name, method_name=_method_name)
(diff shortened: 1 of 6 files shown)


core/src/main/python/wlsdeploy/util/variables.py
```diff
@@ -438,19 +438,33 @@ def _report_token_issue(message_key, method_name, model_context, *args):
...
+
+    matches = _environment_pattern.findall(text)
     for token, key in matches:
-        if key in variables:
-            value = variables[key]
-            text = text.replace(token, value)
+        # log, or throw an exception if key is not found.
+        if not os.environ.has_key(key):
+            _report_token_issue('WLSDPLY-01737', method_name, model_context, key)
+            continue
```
rakillen, line 463: Why not just leave this in the merged model as `@@ENV:...@@` if it can't be resolved, similar to the way `@@PROP` is done here?
I think it's reasonable (in general) to merge models with unresolved ENV or PROP, and log/throw any problems later when the model is used.
The changes to method `_get_merge_match_key` should take care of merging names when the ENV value is matched to a fixed value in the other model (WDT-552).

jshum2479: We can do that.

rakillen (approved):

core/src/main/python/wlsdeploy/util/cla_helper.py
```diff
@@ -305,6 +305,7 @@ def merge_model_files(model_file_value, variable_map=None):
     It may be a single file, or a comma-separated list of files.
     :param variable_map: variables to be used for name resolution, or None
     :param model_file_value: the value specified as a command argument
+    ;param model_context:  model context
```
ddsharpe, line 308: looks like a semi-colon here instead of a colon

rakillen: model_context isn't an argument any more

jshum2479: removed

ddsharpe (approved):

PR #876, opened 2021-04-21
Fix -updateRCUSchemaPassword failed when it has new custom datasource

jshum2479:
This fixes the failure when updating a jrf domain's rcu password but at the same time adding a new datasource.

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/util/rcu_helper.py
+++ b/core/src/main/python/wlsdeploy/tool/util/rcu_helper.py
@@ -56,9 +56,11 @@ class RCUHelper(Deployer):
                 password_location = LocationContext(location)
                 list_path = self.aliases.get_wlst_list_path(location)
                 if not self.wlst_helper.path_exists(list_path):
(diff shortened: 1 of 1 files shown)


rakillen (approved):

CarolynRountree (approved):

core/src/main/python/wlsdeploy/tool/util/rcu_helper.py
```diff
@@ -54,6 +54,12 @@ def update_rcu_password(self):
             for ds_name in ds_names:
                 location = deployer_utils.get_jdbc_driver_params_location(ds_name, self.aliases)
                 password_location = LocationContext(location)
+                list_path = self.aliases.get_wlst_list_path(location)
+                if not self.wlst_helper.path_exists(list_path):
+                    # For update case when a new custom data source has not be persisted
```
ddsharpe, line 59: "data source has not be persisted" => "data source has not been persisted"

PR #893, opened 2021-05-08
switched to using bash array in shell scripts

ddsharpe:
Switched to using bash arrays to cleanup issue with shell not removing quotes for paths with spaces.  Passing value to scriptArgs as $* was causing the double-quote to not get removed.

Also, replaced old-style back-tick with $() syntax.


robertpatrick (approved):

rakillen (approved):

CarolynRountree (approved):

ddsharpe:
obsolete.  Alternate way to accomplish the same thing without using bash merged in #900

PR #897, opened 2021-05-11
Warn if custom security provider in compareModels model comparison

CarolynRountree:
Log a warning message if a custom provider is found in either model being compared with compareModel tool. The custom provider will not be included in the diffed persisted models.

### DIFF
--- a/core/src/test/python/compare_model_test.py
+++ b/core/src/test/python/compare_model_test.py
@@ -125,8 +125,8 @@ class CompareModelTestCase(unittest.TestCase):
     def testCompareModelSecurityConfiguration(self):
         _method_name = 'testCompareModelSecurityConfiguration'
 
-        _new_model_file = self._resources_dir + '/compare_model_modela.yaml'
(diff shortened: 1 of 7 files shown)


rakillen (approved):

core/src/test/python/compare_model_test.py
```diff
@@ -122,6 +122,54 @@ def testCompareModelFull(self):
 
         self.assertEqual(return_code, 0)
 
+    def testCompareModelSecurityConfiguration(self):
+        _method_name = 'testCompareModelSecurityConfiguration'
+
+        _new_model_file = self._resources_dir + '/compare_model_modela.yaml'
```
ddsharpe, line 128: I do not see these yaml files in the source.  Were they meant to be checked in?

CarolynRountree: added. thanks

rakillen (approved):

ddsharpe (approved):

Issue #903, opened 2021-05-25
Removed server glitch in domain servers directory


pwninja:
I am running using WDT for OSB domain installation.
In the model file I am removing default server by using the exclamation mark notation:
```
    Server:
        # Delete the osb server, which is created by default
        '!osb_server1':
```
This gets interpreted properly that this server is removed.
However, in domain's home servers directory I can find a catalog with "!osb_server" name:
```
-bash-4.2$ ls -al servers
total 8
drwxr-x---.  8 ninjainst ninjainst  143 May 25 12:29 .
drwxr-x---. 36 ninjainst ninjainst 4096 May 25 12:30 ..
drwxr-x---.  9 ninjainst ninjainst   95 May 25 12:28 AdminServer
-rw-r-----.  1 ninjainst ninjainst   14 May 25 12:29 AdminServerTag
drwxr-x---. 10 ninjainst ninjainst  108 May 25 12:31 OsbServer01
drwxr-x---. 10 ninjainst ninjainst  108 May 25 12:32 OsbServer02
drwxr-x---.  3 ninjainst ninjainst   22 May 25 12:27 OsbServer03
drwxr-x---.  3 ninjainst ninjainst   22 May 25 12:27 OsbServer04
drwxr-x---.  3 ninjainst ninjainst   22 May 25 12:27 !osb_server1
```
It contains security and boot.properties file:
```
-bash-4.2$ ls -lR servers/\!osb_server1/
servers/!osb_server1/:
total 0
drwxr-x---. 2 ninjainst ninjainst 29 May 25 12:27 security

servers/!osb_server1/security:
total 4
-rw-r-----. 1 ninjainst ninjainst 157 May 25 12:27 boot.properties
```
I think that at some point WDT ignores the exclamation mark and treats this entry as something to be created.

I am using WDT 1.9.11 version.

pwninja:
It turns out that this code part of domain_creator.py doesn't consider if the server name is "is_delete_name" (from model_helper.py):
```
        server_nodes = dictionary_utils.get_dictionary_element(self._topology, SERVER)
        servers = [self._admin_server_name]

        for model_name in server_nodes:
            name = self.wlst_helper.get_quoted_name_for_wlst(model_name)
            servers.append(name)

        admin_username = self.aliases.decrypt_password(admin_username)
        admin_password = self.aliases.decrypt_password(admin_password)
        encrypted_username = encryptionService.encrypt(admin_username)
[...]

PR #905, opened 2021-05-27
fix boot.properties created for deleted server

CarolynRountree:
Fixes issue #903

fix boot.properties being created for deleted server.

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/create/domain_creator.py
+++ b/core/src/main/python/wlsdeploy/tool/create/domain_creator.py
@@ -89,7 +89,8 @@ from wlsdeploy.tool.util.target_helper import TargetHelper
 from wlsdeploy.tool.util.targeting_types import TargetingType
 from wlsdeploy.tool.util.topology_helper import TopologyHelper
 from wlsdeploy.util import dictionary_utils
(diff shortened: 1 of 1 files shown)


rakillen (approved):

core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -89,7 +89,8 @@
 from wlsdeploy.tool.util.targeting_types import TargetingType
 from wlsdeploy.tool.util.topology_helper import TopologyHelper
 from wlsdeploy.util import dictionary_utils
-from wlsdeploy.util import model as model_helper
+from wlsdeploy.util import model as model_assist
```
ddsharpe, line 92: To keep this in sync with other imports throughout the code, can this be changed to 
"import wlsdeploy.util import model" without the "as"?  All of the other imports for model and almost all of the other wlsdeploy.util imports are written that way.

CarolynRountree: I considered, but I figured the original author did not want to name it model for confusion. Just a consideraton

ddsharpe (approved):

Issue #566, opened 2020-03-11
libOVD keystore


dborysenko:
Hi Team,
We've discovered that if one wants to have weblogic 12c domain (with Oracle Service Bus and/or Oracle SOA suite) talk to ldapS authentication provider , the one should setup libOVD keystore (procedure is described in Doc ID 1465228.1). I wonder if it is something that can be implemented in WDT? So libOVD keystore configuration been done during domain creation or post creation step?

ddsharpe:
I have been unable to get traction with the SOA/OSB development team.  Without insight or review from people that know more about libOVD configuration, I am very reluctant to release a feature in this area.  For now, I am archiving this issue until such a time that the SOA/OSB team has time to discuss WDT.

PR #929, opened 2021-06-23
delete removing items from a list attribute

CarolynRountree:
Fixes Jira #566


documentation/1.9/content/concepts/model.md
```diff
@@ -262,6 +262,18 @@ This feature does not apply to named security providers within a realm. These it
 
 This feature cannot be use to undeploy applications or remove libraries.
```
rakillen, line 263: I checked, this is no longer true, apps and libraries can be removed this way.

CarolynRountree: changed to "can"

rakillen (approved):

ddsharpe (approved):

Issue #940, opened 2021-07-13
Error in YAML validator


belfo:
Hello,

i generated a model.yaml using the discovery tool.
After that i run the validateModel on this file but it fail as the name of the AuthenticationProvider contain a &

AuthenticationProvider:
          'Test&Identity Asserter':

Removing the & make the validation ok

Regards

CarolynRountree:
I have never seen discover put a & in the Trust Identity Server name. Could you rerun discover and provide the model and also provide the config.xml? You can mail to carolyn.rountree@oracle.com

belfo:
The name of the Authentication provider contain a & on Weblogic.
In the config it's marked as:
<sec:authentication-provider>
<sec:name>Test&amp;Identity Asserter</sec:name>

So the discover correctly recover it as 
myrealm:
        AuthenticationProvider:
          'Test&Identity Asserter':
            weblogic.management.security.authentication.AuthenticationProvider:

But when i wanted to create an image using the model, or just running the validateModel script, it fail to validate the name.

The YAML syntax in the code don't support a & in the name. The fix just add the char to the syntax. And i was able to create the image / validate the model with the fixed version

belfo:
Merged in #954

PR #941, opened 2021-07-13
Fix issue 940

belfo:
#940 
Add support for & in id name


ddsharpe:
@belfo In order to accept this change, you must follow the contributions process outlined in https://github.com/oracle/weblogic-deploy-tooling/blob/main/CONTRIBUTING.md

PR #947, opened 2021-07-19
Set server template listen port after server listen port

CarolynRountree:
in createDomain, if the server template listen port is 7001, a work around is taken to set the port after the server port as it will not show up in the config.xml if set before.


jshum2479 (approved):

core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -832,6 +835,19 @@ def __create_machines_clusters_and_servers(self, delete_now=True):
         if len(server_nodes) > 0:
             self._create_named_mbeans(SERVER, server_nodes, location, log_created=True, delete_now=delete_now)
 
+        # Work around for bug in WLST where Server Template Listen Port must be set after Server
+        # Listen Port for 7001 in order to show up in the config.xml
+        if len(server_template_nodes) > 0:
+            for template in server_template_nodes:
+                listen_port = dictionary_utils.get_dictionary_element(self._topology[SERVER_TEMPLATE][template], LISTEN_PORT)
```
rakillen, line 842: Does this need a check for None if LISTEN_PORT wasn't in the model?

rakillen (approved):

ddsharpe (approved):

PR #953, opened 2021-07-22
Generate secret file with resolved property value if the original value is tokenized as property

jshum2479:
I think this is complete but want opinion whether this has any impact on existing functionality.  Basically for prepare -target operation, if the field is injectable and has been tokenized with property in it's value.   The prepare should:

1   replace it with @@secret in the model 
2. remove the @@prop: referenced in the variable file
3. the generated secret json/script should show the original resolve property value

This PR addresses (3).   (1,2) already in the main branch.


rakillen (approved):

core/src/main/python/prepare_model.py
```diff
@@ -313,6 +314,16 @@ def walk(self):
             full_model_dictionary = cla_helper.load_model(_program_name, self.model_context, self._aliases,
                                                           "discover", WlstModes.OFFLINE)
 
+            # Just in case the credential cache has @@PROP in the value (that's the model has @@PROP as credential value)
+            # , we use the original variable file to resolve it,
```
ddsharpe, line 318: This leading comma seems misplaced.

ddsharpe (approved):

PR #960, opened 2021-08-04
Avoid overwriting variable values with tokens in prepareModel

rakillen:
Override `_check_tokenized` in credential injector to override other token types.
Remove unused tokens after filter is run, similar to `discoverDomain`.

Extracted `ModelPreparer` from `prepare_model.py` for unit testing.
Use `PrepareException` instead of `CompareException` throughout `prepareModel`.
Added variables methods to check for token-only values.
Create base class for unit tests to provide shared methods.

Internal JIRA WDT-572


robertpatrick (changes requested):
The changes have caused the JDBC URLs in the model files to be improperly escaped in the property files.

In the model file, the JDBC URL is jdbc:mysql://10.0.10.6:3306/tododb.  Prior to these changes, it was copied verbatim into the property file like this:

JDBC.myds.URL=jdbc:mysql://10.0.10.6:3306/tododb

With these changes, it looks like this:

JDBC.myds.URL=jdbc\:mysql\://10.0.10.6\:3306/tododb

robertpatrick (approved):

ddsharpe (approved):

PR #962, opened 2021-08-05
Wdt 575 - compareModel needs to return all original unchanged attributes for app/lib

jshum2479:
We need to return all original unchanged attributes for Application and Library also,  the reason is the delta can be used for updating the domain which can triggers undeploy and deploy again,  without the original values, then after the undeploy, the model used will not have enough information for deploy again.

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/compare/model_comparer.py
+++ b/core/src/main/python/wlsdeploy/tool/compare/model_comparer.py
@@ -444,15 +444,6 @@ class ModelComparer(object):
                     key_value = dictionary_utils.get_element(current_folder, key)
                     if key_value is not None:
                         change_folder[key] = key_value
-            # if change_folder and (SOURCE_PATH not in change_folder):
-            #     print 'i ma here'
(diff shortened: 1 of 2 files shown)


core/src/main/python/wlsdeploy/tool/compare/model_comparer.py
```diff
@@ -428,16 +428,23 @@ def _finalize_folder(self, current_folder, past_folder, change_folder, location)
         if location is not None:
             folder_path = location.get_model_folders()
 
+        app_lib_attributes = [
```
rakillen, line 431: Can we just copy every attribute in current_folder to change_folder that's not already present?
Or is there a reason we have to limit it to this subset of attributes?
It would be nice not to have to list all these attributes in this code.

jshum2479: Yes, I also found logic flaw.

rakillen (approved):

ddsharpe (approved):

PR #980, opened 2021-09-17
Wdt 822 discover roles

### LINKED ISSUE #822: WDT discovery does not get global roles
I have used discover domain tool to introspect a domain. However, the model does not include **WebLogic Global Roles**, is this part of the functionality provided by discover domain? If not, how could I model several global roles? Let us say I want to add several groups into the Admin role. I have seen the example [here](https://github.com/oracle/weblogic-deploy-tooling/blob/master/site/use_cases.md#modeling-weblogic-users-groups-and-roles) and it does not show this specific case.

Thanks a lot in advance.

CarolynRountree:
Fixes #822


core/src/main/python/wlsdeploy/tool/discover/domain_info_discoverer.py
```diff
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2017, 2019, Oracle Corporation and/or its affiliates.  All rights reserved.
+Copyright (c) 2017, 20121, Oracle Corporation and/or its affiliates.
```
robertpatrick, line 2: Typo in the 2021 date.

robertpatrick (changes requested):

core/src/main/java/oracle/weblogic/deploy/aliases/TypeUtils.java
```diff
@@ -270,10 +270,8 @@ public static Character convertToCharacter(String strValue) {
 
     public static Object[] convertToObjectArray(Object value, String strValue, String delimiter)
         throws AliasException {
-        System.out.println("I am in the method" + strValue);
         Object[] result;
         if (Object[].class.isAssignableFrom(value.getClass())) {
-            System.out.println("is assignable " + strValue);
             result = Object[].class.cast(value);
         } else if (value instanceof List) {
             System.out.println("Is instance of List");
```
rakillen, line 277: there's another println here, as long as you're in there...

rakillen (approved):
one comment about println, up to you if you want to remove

ddsharpe (approved):

PR #991, opened 2021-10-25
Online update fix for JIRA 576

jshum2479:
This is trying to address some minor code bugs for online update where we added the library name to the application undeploy this, therefore online update failed.  

At the same time, I found trying to figure out all the dependent applications for the shared library and automate the entire sequence of stop, undeploy, update, deploy, and start problematic because:

1)  Current console does not support this.  If the user tries to update a shared library, all dependent applications must be undeployed first, stopping alone is not enough, otherwise it will failed.

2)  User may only provide a sparse model to update the domain, for example: just the shared library in the model.  If we based on the model, after stopping and undeploying the application.  The sparse model does not have the application for deploying the application again.  This is especially true to WKO MII online update.  

3)  What if the shared library also depends on other shared library?  This creates a transitive situation where we have to keep track of the entire graph of all related applications and libraries which will further complicates the problem. e.g. ordering of undeployment.

4)  In a typical real world environment, user should be careful with the version of the shared library and the application using the shared library.   e.g.  update should update the implementation version of the shared library in the manifest file.  For application that is using the shared library can selectively update the deployment descriptor to picks up the version the shared library.   This way user can select which application to upgrade to the new version of the shared library at the appropriate time.  

5)  This only removes the automatic handling of undeploy / deploy of application that uses shared library when that particular shared library is being updated.   It does not affect other online deployment of applications or shared libraries.  User just have to provide the model that has the entries of  both shared libraries and dependent applications when performing online updates.


robertpatrick (approved):

robertpatrick:
@jshum2479 tests are failing so these must be cleaned up before merging...

jshum2479:
The failure is from previous JRF DB issue.  This is the online update that we agreed to table it for more investigation about the impact,  I added the WIP label, probably revisit early next year to finalize the approach.

ddsharpe:
If/When this moves forward, the copyrights need to be updated to 2022.

documentation/2.0/content/userguide/tools/update.md
```diff
@@ -30,6 +30,22 @@ When running the tool in WLST online mode, the update operation may require serv
...
+  use the new library, this avoids complicated issues like in-place update of shared 
+  library.
+
+- In-place update of shared library online is not supported - if you only 
+  update the library contents without updating the version(s) of the library in the MANIFEST.MF file.  You will get an error 
+  from WebLogic Server indicating the library is referenced by applications and cannot be undeployed. You must undeploy 
+  all applications referencing the shared library first before proceeding; this is the same behavior when using the 
+  WebLogic Server console. Also, a shared library can potentially be referenced by another 
+  shared library module which in turns used by other applications, currently there is no capability within 
+  WebLogic Server to handle automating undeploy and deploy application that uses shared library when the library is
```
ecnjbot, line 46: "undeploy and deploy application" -> "undeploy and deploy of an application"

documentation/2.0/content/userguide/tools/update.md
```diff
@@ -30,6 +30,22 @@ When running the tool in WLST online mode, the update operation may require serv
...
+  library.
+
+- In-place update of shared library online is not supported - if you only 
+  update the library contents without updating the version(s) of the library in the MANIFEST.MF file.  You will get an error 
+  from WebLogic Server indicating the library is referenced by applications and cannot be undeployed. You must undeploy 
+  all applications referencing the shared library first before proceeding; this is the same behavior when using the 
+  WebLogic Server console. Also, a shared library can potentially be referenced by another 
+  shared library module which in turns used by other applications, currently there is no capability within 
+  WebLogic Server to handle automating undeploy and deploy application that uses shared library when the library is 
+  in-place updated.
```
ecnjbot, line 47: "in-place updated." -> "updated in-place."

ddsharpe, line 47: "in-place updated." -> "updated in-place."

documentation/2.0/content/userguide/tools/update.md
ddsharpe, line 46: "undeploy and deploy application" -> "undeploy and deploy of an application"

ddsharpe (approved):

rakillen (approved):

core/src/main/python/wlsdeploy/tool/deploy/applications_deployer.py
```diff
@@ -252,15 +253,22 @@ def __online_deploy_apps_and_libs(self, base_location):
         deployed_app_list = []
         redeploy_app_list = []
```
ddsharpe, line 254: redeploy_app_list does not appear to be used any more.

core/src/main/python/wlsdeploy/tool/deploy/applications_deployer.py
```diff
@@ -272,9 +280,6 @@ def __online_deploy_apps_and_libs(self, base_location):
         self.__deploy_model_libraries(model_shared_libraries, lib_location)
         self.__deploy_model_applications(model_applications, app_location, deployed_app_list)
 
-        for app in redeploy_app_list:
-            self.__redeploy_app(app)
```
ddsharpe, line 276: Does the __redeploy_app method need to be removed also?  Doesn't appear to be used anywhere else.

jshum2479: Yes, I will remove it

PR #993, opened 2021-10-29
Remove usage of array in shared.sh

ddsharpe:
${@:2} is a bash array usage and not supported in all Bourne implementations.


markxnelson (approved):
LGTM, tested on ubuntu 20 with dash

rakillen (approved):

robertpatrick (approved):

PR #1019, opened 2021-12-23
Create validation mode for WKTUI

rakillen:
New mode `wktui` will match `strict`, but allow unresolved secret, environment, and file tokens.
In standalone `validateModel`, list warning and severe messages in summary, without stopping validation.
Standalone `validateModel` in `strict` mode will now flag missing archive entries as errors, and return exit code 2.
See internal Jira WKTUI-215

Use `ValidateConfiguration` to check validation mode.
Some message keys were renumbered.

### DIFF
--- a/core/src/main/python/wlsdeploy/tool/validate/validator.py
+++ b/core/src/main/python/wlsdeploy/tool/validate/validator.py
@@ -822,21 +822,21 @@ class Validator(object):
         #     token to make that explicit in the model.
         #
         if WLSDeployArchive.isPathIntoArchive(path):
+            # If the validate configuration allows unresolved archive references,
+            # log INFO messages identifying missing entries, and allow validation to succeed.
+            # Otherwise, log SEVERE messages that will cause validation to fail.
+            log_method = self._logger.severe
+            if self._validate_configuration.allow_unresolved_archive_references():
+                log_method = _info_logger.info
+
             if self._archive_helper is not None:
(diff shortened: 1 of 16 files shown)


core/src/main/python/wlsdeploy/tool/validate/validator.py
```diff
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2017, 2020, Oracle Corporation and/or its affiliates.
+Copyright (c) 2017, 2021, Oracle Corporation and/or its affiliates.
```
jshum2479, line 2: Should it be 2022?

rakillen: Yes, I'll change all files

core/src/main/python/wlsdeploy/util/model_context.py
```diff
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2017, 2020, Oracle Corporation and/or its affiliates.
+Copyright (c) 2017, 2021, Oracle Corporation and/or its affiliates.
```
jshum2479, line 2: same here for year 2022

rakillen: Updated in last commit

rpatrick00 (approved):

robertpatrick (approved):

ddsharpe (approved):

PR #1022, opened 2022-01-06
document mbean schema jar alternate directory variable

### LINKED ISSUE #946: Deployment of custom mbeans (security providers)
Actually, I just noticed issue #944 sorted, and my interest coincides with that isse in general, but at this time most specifically regarding deployment of custom mbeans (we have several, for both authorization and auditing for example). I figured I'd put this as separate issue so as not to cloud the original.

===

I'm researching how to convert our present domain-creation style to use WDT instead. But at present I'm in a bind :-(

For this issue I've created a repro'able setup; project attached. I have created a nonsense Auditor ('local.ken1.SampleWlsAuditor') that does nothing except printing to stdout on calls. My goal is to model this and deploy it using WDT, during a createDomain. I'm using WDT 1.9.16, Java 11 and WLS14.

Once I started reading up on it, [The Documentation](https://oracle.github.io/weblogic-deploy-tooling/samples/securityproviders-model/#custom-security-providers) is pretty clear - I apparently have to put the mbean jar (and the schema jar) inside the Oracle_Home in order for this to work (and it does). This is really bad IMHO; at least for us.

I have experimented far and wide to break this limitation, but no success.

My problem is first, that our WLS installs are effectively read-only - the Oracle_Home should not be touched by projects using it. The second problem is that even if we managed to relax this, there is simply no good answer as to what would happen when multiple projects use a single Oracle_Home. At best, they would simply contaminate it for everyone (e.g. project A would see securityproviders particular to project B), but at worst they would need to use different versions of the 'same' provider. I can't see any provision to handle that situation (other than installing 'private' Oracle_Home's for every project - while theoretically possible, not very palatable...).

I believe that there must be a mechanism to allow WDT/WLST to find the requisite mbean/schema jars by some alternate mechanism so the Oracle_Home will not need to be touched. The weblogic.alternateTypesDirectory property appears not to impact domain creation (and unclear if that would work for a schema jar also?).

I'm hoping there is a possible workaround for this - or, one can be devised (likely in terms of #944). At present, this is a bit of a deal-breaker for me. I appreciate the suggestion on using WIT, but involving docker containers at this point would complicate the situation too much (we're moving in that direction, but for the vast majority of several thousand projects this is not practical just now).

Any suggestions? Please? :-)

Thanks,

ken1

[sample_wls_auditor.zip](https://github.com/oracle/weblogic-deploy-tooling/files/6825912/sample_wls_auditor.zip)

CarolynRountree:
Fixes #946

Document env variable for alternate location to schema mbean jar.


ddsharpe (approved):

documentation/1.9/content/samples/securityproviders-model.md
```diff
@@ -124,7 +124,7 @@ description: "Special handling and model semantics for WebLogic Server security
...
 
- WebLogic allows you to define an alternate directory other than `WLSERVER/server/lib/mbeantypes` by using the system property `-Dweblogic.alternateTypesDirectory=dir`. In order for the custom provider jars to be loaded correctly by WLST when discovering or configuring a domain, set this system property in the `WLSDEPLOY_PROPERTIES` environment variable.
+ For the mbean jar, WebLogic allows you to define an alternate directory other than `WLSERVER/server/lib/mbeantypes` by using the system property `-Dweblogic.alternateTypesDirectory=dir`. For the WebLogic mbean schema type jar, you can use an alternate location by using -fmwconfig.alternateSchemaDirectory. In order for the custom provider jars to be loaded correctly by WLST when discovering or configuring a domain, set this system property in the `WLSDEPLOY_PROPERTIES` environment variable. Both of the properties take a comma separated list of paths to dirs containing the corresponding type of jar.
```
rakillen, line 127: I think "MBean" is correct - WLS docs and this page seem to use it.

ddsharpe: [oracle@96aea00aff4b oracle]$ oracle_common/common/bin/wlst.sh -fmwconfig.alternateSchemaDirectory

Initializing WebLogic Scripting Tool (WLST) ...

Jython scans all the jar files it can find at first startup. Depending on the system, this process may take a few minutes to complete, and WLST may not return a prompt right away.

Welcome to WebLogic Server Administration Scripting Shell

Type help() for help on available commands


The WebLogic Scripting Tool (WLST) is a command-line 
scripting interface that system administrators and 
operators use to monitor and manage WebLogic Server 
instances and domains.
 
Command line Options are: 
-i	: Runs the script and forces WLST in interactive mode. 
-help	: Prints this help message. 
-skipWLSModuleScanning	: Skips package scanning and caching for WLS modules. 
args	: Arguments can be a script file you would like to execute

ddsharpe: I think this is supposed to be something like:
```
CONFIG_JVM_ARGS="${CONFIG_JVM_ARGS} -Dfmwconfig.alternateSchemaDirectory=dir1[,dir2...]    -Dfmwconfig.alternateTypesDirectory=dir1[,dir2...]
```

ddsharpe (changes requested):

rakillen (approved):

rakillen (commented):
This needs to go into the new 2.0 doc tree now.
@ddsharpe - Just there, or in both places?

ddsharpe (approved):

PR #1049, opened 2022-01-20
Update alias definition documentation

rakillen:


robertpatrick (approved):

rosemarymarano (approved):
I looked at these files locally in Hugo and they appear ok, the only weirdness is a semi-invisible ellipsis at the end of the code block. Not sure if that's important.

PR #1066, opened 2022-01-28
switching to new jgit build number plugin to address error during release process

robertpatrick:
I have verified that this resolves the build number plugin error that was occurring during the release process, forcing us to rebuild the binaries after the release process.


rakillen (approved):

ddsharpe (approved):
needs new 3rd party entries

Issue #1071, opened 2022-02-02
RCU datasources always use derby in a compact domain


httpants:
I'm using version 2.0.0 of weblogic-deploy-tooling that supports the new compact domain topology profile (thanks for adding that by the way). There is an issue when using compact domains if you want to use a non derby database for any of the RCU datasources.

I'm running the following command...
```bash
$ createDomain.cmd -domain_parent target -domain_type OSEF -model_file osefDomain.yml
```
This is my osefDomain.yml template file...

```yaml
domainInfo:
    AdminPassword: welcome1
    ServerStartMode: dev
    RCUDbInfo:
        rcu_prefix: OSEF
        rcu_schema_password: xxxxxxxx
        rcu_db_conn_string: 'localhost:1521/XE'
        rcu_db_user: OSEF_STB
topology:
    Name: DefaultDomain
    AdminServerName: DefaultServer
    Server:
        DefaultServer:
            ListenAddress: localhost
            SSL:
                HostnameVerificationIgnored: true
                TwoWaySSLEnabled: true
                Enabled: true
resources:
    JDBCSystemResource:
        LocalSvcTblDataSource:
            Target: DefaultServer
            JdbcResource:
                JDBCConnectionPoolParams:
                    StatementCacheSize: 10
                    InitialCapacity: 1
                    MinCapacity: 1
                    TestTableName: SQL SELECT 1 FROM DUAL
                    StatementCacheType: LRU
                    MaxCapacity: 15
                JDBCDataSourceParams:
                    GlobalTransactionsProtocol: None
                    JNDIName: jdbc/LocalSvcTblDataSource
                JDBCDriverParams:
                    URL: 'jdbc:oracle:thin:@//localhost:1521/XE'
                    PasswordEncrypted: 'xxxxxxxx'
                    DriverName: oracle.jdbc.OracleDriver
                    Properties:
                        user:
                            Value: OSEF_STB
```
Notice above I specify the LocalSvcTblDataSource datasource to be `jdbc:oracle:thin:@//localhost:1521/XE`.

This is my `wdt_custom_config\typesdefs\OSEF.json` file...

```json
{
    "copyright": "Copyright (c) 2021 Energy Queensland.  All rights reserved.",
    "name": "OSEF",
    "description": "OSEF type domain definitions",
    "topologyProfile": "Compact",
    "versions": {
        "12.2.1.4": "OSEF_12CR2",
        "14.1":     "NOT_SUPPORTED"
    },
[...]

PR #1072, opened 2022-02-03
Wdt=1071 set sample db

CarolynRountree:
Fixes issue #1071

Per conversation, put the useSampleDatabase parameter in the domainInfo. Set option if it is in the domainInfo


robertpatrick (changes requested):
Remove these unrelated logging changes.  I am editing these files so please do not check in any changes at this time, no matter how minor.

core/src/main/resources/oracle/weblogic/deploy/aliases/category_modules/DomainInfo.json
```diff
@@ -11,6 +11,7 @@
         "ServerGroupTargetingLimits": [ {"version": "[10,)", "wlst_mode": "both", "wlst_name": "ServerGroupTargetingLimits", "wlst_path": "WP001", "default_value": null,    "wlst_type": "dict" } ],
         "DynamicClusterServerGroupTargetingLimits": [ {"version": "[12.2.1.1,)", "wlst_mode": "both", "wlst_name": "dynamicClusterServerGroupTargetingLimits", "wlst_path": "WP001", "default_value": null,    "wlst_type": "dict" } ],
         "ServerStartMode":            [ {"version": "[10,)", "wlst_mode": "both", "wlst_name": "ServerStartMode",            "wlst_path": "WP001", "default_value": null,    "wlst_type": "string" } ],
+        "UseSampleDatabase":          [ {"version": "[10,)", "wlst_mode": "both", "wlst_name": "UseSampleDatabase",          "wlst_path": "WP001", "default_value": null,    "wlst_type": "string" } ],
```
robertpatrick, line 14: I do not believe that this is applicable to 10.3.x or even 12.1.x. I think the compact topology profile was added in 12.2.x so I suspect that this option isn't really applicable before 12.2.1, though we should verify.

robertpatrick: The JRF-Compact typedef file is currently marking anything before 12.2.1 as NOT_SUPPORTED

robertpatrick: Also, I don't think this is applicable to online since it controls which database a JRF (or upper stack) domain's DataSources point to...which is only applicable prior to the domain being created.

CarolynRountree: there are a few that way, can you setOptions after createDomain? What about ServerStartMode

ddsharpe (approved):

rakillen (approved):

robertpatrick (approved):

PR #1076, opened 2022-02-05
Add new target environments wko-dii, wko-pv, vz-dii, vz-pv

rakillen:
Also, move exposed filter methods to internal filters.


core/src/main/python/wlsdeploy/tool/util/filter_helper.py
```diff
@@ -18,9 +18,16 @@
 TARGET_CONFIG_TOKEN = '@@TARGET_CONFIG_DIR@@'
```
robertpatrick, line 18: Take out the "All rights reserved." sentence from the copyright.

robertpatrick (approved):

jshum2479 (commented):
For PV cases, the domain resource yaml should have something like this... 
```
serverPod:
      volumes:
      - name: weblogic-domain-storage-volume
        persistentVolumeClaim:
           claimName: sample-domain1-pvc-rwm1
      volumeMounts:
      - mountPath: /shared
        name: weblogic-domain-storage-volume
```        

name in volumeMount and Volume pair must match 
claimName is the k8s pvc name
mountPath is user desired  and one of them should be used by the domainHome's starting part.

ddsharpe (approved):

jshum2479 (approved):

PR #1087, opened 2022-02-10
Wdt 608 undo adjudicator code

CarolynRountree:
fixes Jira 608

Changing adjudicator in versions 12.2.1.3 and below encounters a WLST problem with the xsi type in the config.xml. Have opened  WLST bug 33847230. put changes in to handle version specific problems with the adjudicator.


robertpatrick:
The adjudicator has been in WLS for a long time so are we sure it doesn't work in 10.3 and 12.1.x?

rakillen (approved):

CarolynRountree:
The tool has had messages in it about the adjudicator since securityconfigurations went in I suppose. I undid them when testing in 12.2.1.4 as I did not remember why it did not work. 1.9.18. I reapplied those changes along with a version check since 12.2.1.4 is working.

ddsharpe (approved):

PR #1109, opened 2022-03-24
Wdt 609 ssl db connection

CarolynRountree:
RCU SSL DB in WDT createDomain. I used the atp as a template and the code follows the atp code.


rakillen (approved):

rakillen (approved):
There are some conflicts with sonar changes, but 👍 once that passes

core/src/main/python/wlsdeploy/tool/create/domain_creator.py
```diff
@@ -1004,6 +1014,43 @@ def __retrieve_atp_rcudbinfo(self, rcu_db_info, checkAdminPwd=False):
...
         return tns_admin, rcu_database, keystore_pwd, truststore_pwd
 
+    def __retrieve_ssl_rcudbinfo(self, rcu_db_info, checkAdminPwd=False):
+        """
+        Check and return ssl connection info and make sure ssl rcudb info is complete
+        :raises: CreateException: if an error occurs
+        """
+        _method_name = '__retrieve_ssl_rcudbinfo'
+
+        tns_admin = rcu_db_info.get_ssl_tns_admin()
```
jshum2479, line 1024: Is tns_admin required field ?   If yes, we should probably check the truststore field value and the combination of them pointing to an existing file

jshum2479:
Will there be any documentation update?

core/src/main/java/oracle/weblogic/deploy/create/RCURunner.java
```diff
@@ -174,6 +176,56 @@ public static RCURunner createAtpRunner(String domainType, String oracleHome, St
...
+        if (trustStorePassword != null && trustStorePassword != "None") {
+            sslArgs.append(",javax.net.ssl.trustStorePassword="+ trustStorePassword);
+        }
+        sslArgs.append(",javax.net.ssl.keyStore=");
+        sslArgs.append(tnsAdmin + "/" + keyStore);
+        sslArgs.append(",javax.net.ssl.keyStoreType=" + keyStoreType);
+        if (keyStorePassword != null && keyStorePassword != "None") {
+            sslArgs.append(",javax.net.ssl.keyStorePassword="+ keyStorePassword);
+        }
+        sslArgs.append(",oracle.net.ssl_server_dn_match=false");
```
jshum2479, line 223: do we need to make ssl_server_dn_match configurable ?

CarolynRountree: It was in a different branch and was already checked into the documentation main.

CarolynRountree: I'm not sure we do, from what I have seen everywhere, it is false.

jshum2479: Yes, I know in ATP we hard coded to false and probably we need to make it configurable - just in case user complains against it (most internal users have not much experience anyway, I don't know what it is until I read the doc) -  we can make it default to false?   I think this is an extra level of security to make sure talking to the right db server.  

From the documentation,

https://www.oracle.com/technetwork/database/enterprise-edition/wp-oracle-jdbc-thin-ssl-130128.pdf

The following property also needs to be used to force the JDBC Thin driver to verify the server’s DN:
props.setProperty("oracle.net.ssl_server_dn_match", "true");

core/src/main/python/wlsdeploy/tool/create/domain_creator.py
CarolynRountree: Added thetruststore check to the tnsnames check

core/src/main/java/oracle/weblogic/deploy/create/RCURunner.java
CarolynRountree: documentation was in another branch and it got merged already into documentation.

CarolynRountree: from what I read online it should always be false. I might be wrong.

jshum2479: Most example doesn't want to have the hassle for user to handle error.  The property exists for a reason, even though user may usually set it to false,  why not provide a option for user?

https://blogs.oracle.com/developers/post/ssl-connection-to-oracle-db-using-jdbc-tlsv12-jks-or-oracle-wallets-122-and-lower

https://docs.oracle.com/en/middleware/fusion-middleware/data-integrator/12.2.1.4/odi-marketplace/known-issues-and-workarounds.html

jshum2479: Note: I am saying setting it to true, I am just asking to make it a configurable property

CarolynRountree: sorry, I replied to your first comment twice. changing to a property

rjeberhard:
I reviewed the security hotspots and don't have any concerns. Derek, you should review these too please.

@CarolynRountree, can you please clean up the code smells? I think that those do look valid.

robertpatrick (approved):

jshum2479 (approved):

PR #1127, opened 2022-05-16
remote processing of discover

CarolynRountree:
Fixes Jira WDT-631

This allows online discovery of a remote domain. Added new command -remote that will put the archive name in the model file instead of local name.

Prints information out to STDOUT


core/src/main/python/discover.py
```diff
@@ -224,8 +245,17 @@ def __discover(model_context, aliases, credential_injector, helper):
                                                         ae.getLocalizedMessage(), error=ae)
         __logger.throwing(ex, class_name=_class_name, method_name=_method_name)
         raise ex
-
     __disconnect_domain(helper)
+
+    if model_context.is_remote():
+        print 'Discovered Remote Domain - showing files to collect for archive'
```
ddsharpe, line 251: Why are you using "print" instead of __logger?

installer/src/main/bin/discoverDomain.sh
```diff
@@ -61,6 +61,10 @@ usage() {
   echo "        archive_file    - the path to the archive file to use"
   echo ""
   echo "        skip_archive    - Do not generate the archive file. The archive file command will be ignored"
+  echo "                          The file references in the model will be the local domain file names"
+  echo ""
+  echo "        remote          - Online only. Discover the remote domain. Do not generate the archive file. "
+  echo "                          The file references in the model will generated as archive file names for ease of use"
```
rakillen, line 67: The .sh description and .cmd descriptions are different.

core/src/main/python/discover.py
CarolynRountree: It was requested by Robert to not mix in with Summary. As a separate piece of stdout

robertpatrick: In fact, you cannot use the logger.  Otherwise, it gets mixed up with all other log messages.  The Summary Logger only supports Warning/Error levels so we cannot use that either.

core/src/main/java/oracle/weblogic/deploy/util/WLSDeployArchive.java
```diff
@@ -581,7 +609,15 @@ public String addApplication(File appPath) throws WLSDeployArchiveIOException {
 
         LOGGER.entering(CLASS, METHOD, appPath);
         String newName = appPath.getPath();
-        if (hasArchive) {
+        String localName = newName;
```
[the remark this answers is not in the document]
CarolynRountree: done

rakillen (approved):

core/src/main/java/oracle/weblogic/deploy/util/WLSDeployArchive.java
```diff
@@ -571,20 +599,27 @@ public String getFileHash(String path) throws WLSDeployArchiveIOException {
...
+        File filePath = new File(appPath);
+        String newName = appPath;
+        if (isRemote) {
+            newName = addItemToZip(ARCHIVE_APPS_TARGET_DIR, filePath);
+            addToRemoteMap(appPath, newName, ArchiveEntryType.APPLICATIONS.name());
+        }
+        else if (hasArchive) {
+            validateExistingFile(filePath, "appPath", getArchiveFileName(), METHOD, true);
+
+            newName = addItemToZip(ARCHIVE_APPS_TARGET_DIR, filePath);
```
robertpatrick, line 622: As discussed, this feels strange.  We are calling `addXXX()` methods on a class named `WLSDeployArchiveFile` even when we are not really adding anything to the archive file.  We are also calling `addItemToZip`, when there is no zip file.  To me, this makes the code harder to understand.  It seems like the abstraction level is wrong.  I would think that the `WLSDeployArchiveFile` shouldn't need to understand how to behave when we are not using an archive file.  Also, this abstraction will likely need to be refactored once we can gather remote files since it is possible that we fail to gather a remote file (so we need to add it to the list of user actions) while we successfully gather other files.

core/src/main/java/oracle/weblogic/deploy/util/WLSDeployArchive.java
```diff
@@ -1618,6 +1735,13 @@ private static void validateExistingFile(File file, String argName, String fileN
         LOGGER.exiting(CLASS, METHOD);
     }
 
+    private static void addToRemoteMap(String localName, String newName, String type) {
+        Map<String, String> remoteMap = new HashMap<>();
+        remoteMap.put(REMOTE_TYPE, type);
+        remoteMap.put(REMOTE_ARCHIVE_DIR, newName);
+        remoteList.put(localName, remoteMap);
+    }
```
robertpatrick, line 1743: Again, this feels wrong.  Why should the `WLSDeployArchiveFile` keep track of this map of user actions?  This is a discovery-only map for reporting discovery results to the user.  It does not really belong here.

robertpatrick (changes requested):

core/src/main/java/oracle/weblogic/deploy/util/WLSDeployArchive.java
rakillen: I agree that this adds a lot of complexity to `WLSDeployArchiveFile`, and will really blow up once we try the SSH stuff. Maybe an accessor class with matching `add*()` functions could handle the extra decision making and reporting, and only call `WLSDeployArchiveFile` methods when we're really adding. That would minimize the impact to existing discover code.
The accessor class methods could also take string arguments and pass java.io.File objects to the corresponding `WLSDeployArchiveFile` methods. Maybe `ArchiveUpdater`?

rakillen (commented):
See comments about changes to WLSDeployArchive

rakillen (changes requested):
See comments about changes to WLSDeployArchive

robertpatrick (approved):

ddsharpe (approved):

PR #1158, opened 2022-07-06
Build simple-app.war correctly for system test

rakillen:
*** Confirmed that simple-app.war and archive.zip are correct in the pr-merge workspace ***

Exit if any of the `cd` commands fail.

Internal Jira WDT-656


robertpatrick (approved):
Approving as an intermediate step to refactoring system-test to be a multi-module project

PR #1246, opened 2022-11-04
Add information about wko4 targets

rakillen:


documentation/2.0/content/userguide/target_env.md
```diff
@@ -130,6 +132,22 @@ kubernetes:
 ```
 These fields will override the values in the output file, and the file would be rewritten with the revised values. List values in the model will be combined with existing values in the output file. For example, if `my-cluster` was in the original output file, the model content for `my-cluster` would be merged with it, overriding the `replicas` value. If `my-cluster` was not in the original output file, it would be added to the list of clusters.
 
+When creating a resource file for WebLogic Kubernetes Operator 4.0, the `kubernetes` section of the model has to be structured differently. This is because there are multiple documents in the resource file for version 4.0. This example uses `domain` and `cluster` folders that will merge with the corresponding documents in the resource file:
```
robertpatrick, line 135: WebLogic Kubernetes Operator 4.0 => WebLogic Kubernetes Operator 4.0.0+

robertpatrick: Do we also need to state that the `wko` targets are for WKO 3.x just to be clear?

documentation/2.0/content/userguide/target_env.md
```diff
@@ -42,9 +42,11 @@ These target environment configurations are included in the WebLogic Deploy Tool
 
 You can use these targets to customize the model and create a domain resource file for use with WebLogic Kubernetes Operator. There are three targets for specific [domain home source types](https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/choosing-a-model/):
 
-- `wko` for [Model in Image](https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/model-in-image/) deployments
-- `wko-dii` for Domain in Image deployments
-- `wko-pv` for Domain in PV deployments
+- `wko` and `wko4` for [Model in Image](https://oracle.github.io/weblogic-kubernetes-operator/userguide/managing-domains/model-in-image/) deployments
```
rosemarymarano, line 45: In WKO 4.0, the link to the Model in Image docs has been changed to https://oracle.github.io/weblogic-kubernetes-operator/managing-domains/model-in-image/

rosemarymarano (approved):
One change needed.

documentation/2.0/content/userguide/target_env.md
rakillen: Updated, including changes for broken links elsewhere.

rosemarymarano (approved):
Thank you!!!

PR #1284, opened 2022-11-21
fix restfulmanagementservices attributes

CarolynRountree:
Correct errors for RestfulManagementService


robertpatrick (changes requested):
None of these attributes are available in WLST offline in 12.2.1.4.x

robertpatrick (approved):

robertpatrick (approved):

PR #1289, opened 2022-11-22
hostingserver ends at 12.2.1.4

CarolynRountree:
migratabletarget extends singletonservice. HostingServer ends at 12.2.1.4 as in singletonservice


robertpatrick (changes requested):
HostingServer is available in later versions so this change seems bad...

PR #1355, opened 2023-01-12
Support remote app deploy

jshum2479:
This PR adds support for remote updating a domain and deploying applications.  I also refactored some of the sonar complexities warnings in application_deployer.


core/src/main/python/wlsdeploy/tool/util/wlst_helper.py
```diff
@@ -1361,6 +1361,45 @@ def server_config(self):
...
+            self.server_config()
+            self.cd('/')
+            name = self.get('Name')
+        except self.__load_global('WLSTException'), e:
+            pwe = exception_helper.create_exception(self.__exception_type, 'WLSDPLY-00065',
+                                                    _format_exception(e), error=e)
+            self.__logger.throwing(class_name=self.__class_name, method_name=_method_name, error=pwe)
+            raise pwe
+        return name
+        self.__logger.exiting(class_name=self.__class_name, method_name=_method_name)
```
ddsharpe, line 1381: Should this logging call be before the return?

ddsharpe:
Please double check your copyrights.  core/src/main/python/wlsdeploy/tool/deploy/applications_deployer.py needs to be updated to 2023, not sure about others

robertpatrick (changes requested):
Please remove WDT 3.0 doc changes from the PR and put into the branch for PR #1340

core/src/main/java/oracle/weblogic/deploy/util/WLSDeployArchive.java
```diff
@@ -495,6 +495,52 @@ public boolean containsFileOrPath(String path) throws WLSDeployArchiveIOExceptio
...
+        return result;
+    }
+
+    /** Check whether the path should be used for remote update
+     *
+     * @param path
+     * @return true if the path should not be used for remote update
+     * @throws IllegalArgumentException    if the path is null or empty
+     */
+    public boolean isRemoteUpdateDomainForbiddenPath(String path)  throws IllegalArgumentException  {
```
robertpatrick, line 528: I'm not sure that this is the best place for this logic.  In my opinion, the archive file shouldn't know anything about remote mode.  Can't we put this into archive_helper.py?

core/src/main/python/deploy.py
```diff
@@ -41,12 +41,12 @@
 
 __required_arguments = [
     CommandLineArgUtil.ORACLE_HOME_SWITCH,
-    CommandLineArgUtil.DOMAIN_HOME_SWITCH,
     CommandLineArgUtil.MODEL_FILE_SWITCH
 ]
 
 __optional_arguments = [
     # Used by shell script to locate WLST
+    CommandLineArgUtil.DOMAIN_HOME_SWITCH,
```
robertpatrick, line 49: Is there ever a case where the model for a remote deploy operation might have a domain home token?  If so, what happens if the user did not provide the domain home path on the remote machine?

core/src/main/python/update.py
```diff
@@ -44,12 +46,12 @@
 
 __required_arguments = [
     CommandLineArgUtil.ORACLE_HOME_SWITCH,
-    CommandLineArgUtil.DOMAIN_HOME_SWITCH,
```
robertpatrick, line 47: Same questions as for Deploy.  For discoverDomain -remote, we still require the user to provide -domain_home <remote-domain-home-path> to support tokenization.  I would think we might need this for at least updateDomain -remote if not all remote operations.

robertpatrick (commented):
Comments inline

core/src/main/python/wlsdeploy/tool/util/wlst_helper.py
jshum2479: Fixed

robertpatrick (approved):

